PR #19, opened 2021-11-24
extract_tools.sh: detect LFS installation + fix pull detection

houqp:
* detect git lfs installation
* detect git lfs objects that need to be pulled using `git lfs ls-files` command. `! -f` doesn't work because it returns true for lfs pointer files as well.

### DIFF
--- a/tools/extract_tools.sh
+++ b/tools/extract_tools.sh
@@ -2,6 +2,11 @@
 
 echo "Extracting tools..."
 
+git lfs &> /dev/null || {
+    echo "ERROR: git lfs not installed"
+    exit 1
+}
+
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null && pwd)"
 THIS_SCRIPT="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)/$(basename "${BASH_SOURCE[0]}")"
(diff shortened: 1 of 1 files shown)


tools/extract_tools.sh
```diff
@@ -2,6 +2,11 @@
 
 echo "Extracting tools..."
 
+git lfs &> /dev/null || {
+    echo "ERROR: git lfs not installed"
+    exit 1
```
adeebshihadeh, line 7:
```suggestion
  echo "ERROR: git lfs not installed"
  exit 1
```

tools/extract_tools.sh
```diff
@@ -13,15 +18,14 @@ GOOGLE_GCC_4_9=aarch64-linux-android-4.9
...
-  git lfs pull
-  cd $DIR
-fi
+# grep for `-`, which stands for LFS pointer
+git lfs ls-files | awk '{print $2}' | grep "-" &>/dev/null && {
+    echo "Pulling git lfs objects..."
+    cd $ROOT
+    git lfs install
+    git lfs pull
+    cd $DIR
```
adeebshihadeh, line 27:
```suggestion
  echo "Pulling git lfs objects..."
  cd $ROOT
  git lfs install
  git lfs pull
  cd $DIR
```

PR #29, opened 2021-12-24
Support build_system.sh on M1 Mac

kfatehi:
This PR only focuses on build_system.sh

# build_system.sh

The existing check (uname -p) does not work. the check I added (uname -m) might be sufficient for whatever environment warranted the addition of the first, but for now, this PR adds a check for M1 which works.

```
keyvan@Keyvans-MacBook-Air ~ % uname -p
arm
keyvan@Keyvans-MacBook-Air ~ % uname -m
arm64
```

The actual run of `build_system` ended for me on:

```
Creating empty filesystem
./build_system.sh: line 41: fallocate: command not found
```

So then I made the next commit which creates the empty filesystem inside the docker container.

Here I also removed the /dev/null piping so that in case the thing is giving feedback or asking if you want to overwrite the user isn't stuck wondering what's going on.

Then we get stuck at:

```
Mounting empty filesystem
umount: illegal option -- l
usage: umount [-fv] [-t fstypelist] special | node
       umount -a[fv] [-h host] [-t fstypelist]
mount: You must specify a filesystem type with -t.
```

Because the BSD/Mac umount does not have a -l (show labels). It's at this point I realized probably the rest of the work needs to happen in a container anyway because we lack most of these subsequent disk mounting commands....

So that is what I did in the remainder of the script and it seems to have worked.

image

### DIFF
--- a/build_system.sh
+++ b/build_system.sh
@@ -38,8 +38,8 @@ docker build -f Dockerfile.agnos -t agnos-builder $DIR
 
 # Create filesystem ext4 image
 echo "Creating empty filesystem"
-fallocate -l $ROOTFS_IMAGE_SIZE $ROOTFS_IMAGE
-mkfs.ext4 $ROOTFS_IMAGE > /dev/null
+docker run --rm --mount type=bind,source=$BUILD_DIR,target=$BUILD_DIR -it agnos-builder fallocate -l $ROOTFS_IMAGE_SIZE $ROOTFS_IMAGE
+docker run --rm --mount type=bind,source=$BUILD_DIR,target=$BUILD_DIR -it agnos-builder mkfs.ext4 $ROOTFS_IMAGE
 
 # Mount filesystem
 echo "Mounting empty filesystem"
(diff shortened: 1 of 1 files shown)


robbederks:
Are you sure the image works? 4.3MB (in your screenshot) seems a bit small for the system partition 😅

kfatehi:
> Are you sure the image works? 4.3MB (in your screenshot) seems a bit small for the system partition 😅

Haha no definitely not. It takes around a day or two to build on my 16 core amd64 machine. I turned off `buildkit` and now it's taking much longer so we'll see, i'll post results once it's done in, im assuming, a day or two.

kfatehi:
Updated the screenshot which now shows proper size. Should be good to test, but I don't want to mess up my current tici so I have not tried to flash it and find out.

kfatehi:
Here's a build on linux/amd64 on this branch. Not sure why there is a size difference.

image

adeebshihadeh:
M1 should be supported now. Did a bunch of the production AGNOS builds with a mac mini.

PR #45, opened 2022-02-07
iptables: block incoming wwan0 traffic

pd0wm:


pd0wm:
@adeebshihadeh what do you think of this? Should be a bit less hacky.

PR #102, opened 2023-01-02
XBL + ABL updates

adeebshihadeh:
- [x] fix no touch after update until cold boot
- [x] fix broken serial `msm_geni_serial 88c000.qcom,qup_uart: msm_geni_serial_startup: Invalid FW 3 loaded.`
- [x] figure out minimal set of partitions needed (aop, devcfg, etc?)
- [x] test backwards compatibility with old aop, devcfg, etc.
- [x] test booting up with bad XBL and ABL
- [x] test display init fix at cold and hot temps

### DIFF
--- a/build_kernel.sh
+++ b/build_kernel.sh
@@ -45,7 +45,7 @@ cp $DIR/agnos-kernel-sdm845/out/arch/arm64/boot/Image.gz-dtb .
 $TOOLS/mkbootimg \
   --kernel Image.gz-dtb \
   --ramdisk /dev/null \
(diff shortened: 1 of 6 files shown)


robbederks:
Couldn't reproduce the no-touch until cold boot issue, @adeebshihadeh mind trying it out again? maybe I'm missing something.

If this is fine, and we believe that we've made all the necessary changes, I think we might be good to merge and update.

adeebshihadeh:
Yes, I'll try to reproduce. I'd like to keep this open (or at least wait to ship in openpilot) until we understand the boot issues in cold weather.

adeebshihadeh:
I couldn't repro, even on the old versions. I think we can merge once these two are done:
- [ ] #27
- [ ] https://github.com/commaai/edk2_tici/pull/11

PR #109, opened 2023-02-07
setup improvements

incognitojam:
built on c3 from openpilot master, including these changes
- https://github.com/commaai/openpilot/pull/27235
- https://github.com/commaai/openpilot/pull/27237
- https://github.com/commaai/openpilot/pull/27241
- https://github.com/commaai/openpilot/pull/27236
- https://github.com/commaai/openpilot/pull/27426
- https://github.com/commaai/openpilot/pull/27429

see also 15391fda616b2f6cef7f34d6a5c9808eb3863084


adeebshihadeh (changes requested):
Remaining TODOs
- [ ] write setup URL somewhere, such as `/tmp/setup_url`
- [ ] show a better error when we didn't download a valid installer binary
- [ ] add a test plan to `internal/README.md` for the setup that hits all states
- [ ] go through setup tests

PR #134, opened 2023-05-06
weston: go into idle state

adeebshihadeh:


andiradulescu:
Still relevant? Should we better create an issue for it and close the PR?

adeebshihadeh:
it's still relevant, but feel free to move this to an issue if you'd like

andiradulescu:
done, should we close it then? when the time comes (mainline & weston update), we can make the new weston go to idle.

PR #139, opened 2023-05-15
alternate images compressed with gzip

incognitojam:
system.img 3.7GB
system.img.gz 1.1GB
system.img.xz 714MB

### DIFF
--- a/internal/ota_push.sh
+++ b/internal/ota_push.sh
@@ -34,7 +34,7 @@ process_file() {
     local FILE_NAME="$NAME-$HASH.img.$EXT"
     local CLOUD_PATH="https://$DATA_ACCOUNT.blob.core.windows.net/$DATA_CONTAINER/$FILE_NAME"
 
-    echo "Copying $NAME to the cloud..."
+    echo "Copying $NAME $EXT to the cloud..."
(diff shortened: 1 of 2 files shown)


incognitojam:
flash.comma.ai is using the gzip images

adeebshihadeh:
Oh, I think we should take another stab at using bz2 when we can.

internal/ota_push.sh
```diff
@@ -29,12 +29,15 @@ fi
...
-  local CLOUD_PATH="https://$DATA_ACCOUNT.blob.core.windows.net/$DATA_CONTAINER/$FILE_NAME"
 
-  echo "Copying $NAME to the cloud..."
-  azcopy cp --overwrite=false $OTA_DIR/$FILE_NAME "$CLOUD_PATH?$DATA_SAS_TOKEN"
-  echo "  $CLOUD_PATH"
+  for EXT in xz gz; do
+    local FILE_NAME="$NAME-$HASH.img.$EXT"
+    local CLOUD_PATH="https://$DATA_ACCOUNT.blob.core.windows.net/$DATA_CONTAINER/$FILE_NAME"
+
+    echo "Copying $NAME to the cloud..."
```
incognitojam, line 37:
```suggestion
    echo "Copying $NAME $EXT to the cloud..."
```

PR #151, opened 2023-06-28
Wait for partitions before starting rmt_storage service

robbederks:
The issue this fixes is that `rmt_storage` needs to have these symlinks existing before starting, which are populated by udev and `/etc/udev/scripts/automountsdcard.sh`:

- /dev/block/bootdevice/by-name/modemst1
- /dev/block/bootdevice/by-name/modemst2
- /dev/block/bootdevice/by-name/fsc
- /dev/block/bootdevice/by-name/fsg

Not sure if there's a cleaner way than this, but couldn't figure out how to easily wait for a file to exist before starting a service other than this bash script called in `ExecStartPre=`


adeebshihadeh:
LGTM

PR #164, opened 2023-08-08
Start sshd early

robbederks:
Having both `After=multi-user.target` and `WantedBy=multi-user.target` doesn't sound like a good idea:
https://unix.stackexchange.com/questions/693485/after-multi-user-target-and-others-not-working-in-a-systemd-service

Also, the rescue target is nicer, since that make s sure sshd is up in case any of the runlevel 2 stuff fails.

### DIFF
--- a/userspace/files/ssh-param-watcher.service
+++ b/userspace/files/ssh-param-watcher.service
@@ -1,10 +1,9 @@
 [Unit]
 Description=SSH param watcher
-After=multi-user.target
 
 [Service]
 Type=oneshot
 ExecStart=/usr/comma/set_ssh.sh
 
 [Install]
-WantedBy=multi-user.target
+WantedBy=rescue.target


andiradulescu:
Relating to https://github.com/commaai/agnos-builder/pull/191#issuecomment-1901232894 I have one question. Doesn't `rescue.target` make this run only in "rescue mode" ?
```diff
-WantedBy=multi-user.target
+WantedBy=rescue.target
```

Reading your link above, the first [response](https://unix.stackexchange.com/a/693533/321174) says `"My guess is to add After=network.target if your only issue is with networking."` and I think this suits us better since, from my point of view, the only point of having `ssh` is if we have `networking`.
```diff
-After=multi-user.target
+After=network.target
```

As example, looking at the default `ssh.service` in Ubuntu
```
[Unit]
Description=OpenBSD Secure Shell server
Documentation=man:sshd(8) man:sshd_config(5)
After=network.target auditd.service
ConditionPathExists=!/etc/ssh/sshd_not_to_be_run

[Service]
EnvironmentFile=-/etc/default/ssh
ExecStartPre=/usr/sbin/sshd -t
ExecStart=/usr/sbin/sshd -D $SSHD_OPTS
ExecReload=/usr/sbin/sshd -t
ExecReload=/bin/kill -HUP $MAINPID
KillMode=process
Restart=on-failure
RestartPreventExitStatus=255
Type=notify
RuntimeDirectory=sshd
RuntimeDirectoryMode=0755

[Install]
WantedBy=multi-user.target
Alias=sshd.service
```
the configuration is:
```
After=network.target auditd.service
WantedBy=multi-user.target
```

What do you think?

PR #165, opened 2023-08-11
flash `extra.json` manifest

incognitojam:


internal/extra_push.sh
```diff
@@ -0,0 +1,24 @@
...
+
+source $DIR/upload.sh
+
+# Liftoff!
+for NAME in $(cat $EXTRA_JSON | jq -r '.[] | .name'); do
+  local HASH_RAW=$(cat $EXTRA_JSON | jq -r ".[] | select(.name == \"$NAME\") | .hash_raw")
+  local SPARSE=$(cat $EXTRA_JSON | jq -r ".[] | select(.name == \"$NAME\") | .sparse")
+
+  if [ "$SPARSE" == "true" ]; then
+    upload_file "$NAME-$HASH_RAW-optimized.img.gz"
```
adeebshihadeh, line 18: optimized is kind of a misnomer here - it's just "special". maybe call it "skipped-chunks"

internal/package_extra.sh
```diff
@@ -0,0 +1,120 @@
...
+OTA_OUTPUT_DIR="$OUTPUT_DIR/ota"
+TOOLS_DIR="$DIR/../tools"
+
+AGNOS_UPDATE_URL=${AGNOS_UPDATE_URL:-https://commadist.azureedge.net/agnosupdate}
+AGNOS_STAGING_UPDATE_URL=${AGNOS_STAGING_UPDATE_URL:-https://commadist.azureedge.net/agnosupdate-staging}
+OTA_JSON="$OTA_OUTPUT_DIR/ota.json"
+EXTRA_JSON="$OTA_OUTPUT_DIR/extra.json"
+EXTRA_STAGING_JSON="$OTA_OUTPUT_DIR/extra-staging.json"
+
+process_file() {
```
adeebshihadeh, line 19: seems like this file has a ton of duplication

adeebshihadeh (commented):
Where does the extra manifest live?

PR #166, opened 2023-08-11
swap verizon apn order and add mcc/mnc from gnome

jbschooley:
Verizon wasn’t connecting without setting the apn in openpilot, and took too long to connect, changing the apn order fixes that

mcc/mnc additions taken from here https://github.com/GNOME/mobile-broadband-provider-info/blob/main/serviceproviders.xml


adeebshihadeh:
The order swapping doesn't seem like a real fix. Does this break the lower ones now?

jbschooley:
The other APN is vzwims, which is used for MMS. If you try to connect to Verizon with a data sim with this APN it won’t connect. It seems like when the C3 boots, it just tries to connect with the first APN from the list (then openpilot starts and switches it to the user configured one).

vzwinternet is the correct APN for all Verizon data plans as far as I know, and some MVNOs (at least one user with xfinity mobile used it). Switching the order seems to be making a big difference for me, it connects much faster and connects even if openpilot doesn’t start. Also was done previously for blue prime https://github.com/commaai/agnos-builder/commit/48660ab1c7f8377cf68a510ae8d48fe698d6e6bc

andiradulescu:
Closing stalled PR. Feel free to reopen if still relevant.

jbschooley:
Still an issue. I still have to swap in my modified serviceproviders.xml every time agnos updates.

andiradulescu:
Thanks for the reply! I'll see what I can do.

jbschooley:
Is there really nobody else using verizon with their comma?

PR #167, opened 2023-08-14
optimise system image using simg2dontcare

incognitojam:
Generate an additional "optimized" image for any sparse partition. Adds `url_optimized` and `hash_optimized` properties to manifest.

As the optimized image requires the partition to be erased first, the original image is preserved and still uploaded so that we don't break existing flash/openpilot updater.

split out from https://github.com/commaai/agnos-builder/pull/165

### DIFF
--- a/internal/package_ota.sh
+++ b/internal/package_ota.sh
@@ -72,7 +72,7 @@ EOF
 
   if [ "$SPARSE" == "true" ]; then
     echo "Optimizing $NAME..."
-    local OPTIMIZED_FILE=/tmp/$NAME-optimized.img
+    local OPTIMIZED_FILE=$TMP_DIR/$NAME-optimized.img
     $TOOLS_DIR/simg2dontcare.py $FILE $OPTIMIZED_FILE
 
     echo "Hashing optimized $NAME..."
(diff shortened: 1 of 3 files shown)


adeebshihadeh:
Once this is merged, can you also clean up provisioning to use this?

internal/package_ota.sh
```diff
@@ -70,6 +70,31 @@ EOF
     "url": "$AGNOS_STAGING_UPDATE_URL/$FILENAME"
 EOF
 
+  if [ "$SPARSE" == "true" ]; then
```
adeebshihadeh, line 73: I thought we talked about making this a separate script that takes an agnos manifest?

incognitojam: I can still write that script (to be able to process the existing manifest?), but I wanted to check whether we need to have the output in a separate manifest or not - I didn't know what that helped with

adeebshihadeh: it has to be a different manifest since openpilot's agnos updater flashes all the entries in the manifest

incognitojam: do we want openpilot to use this image eventually?

adeebshihadeh: no

incognitojam:
I don't think this needs to be split out anymore https://github.com/commaai/agnos-builder/pull/165/files

PR #170, opened 2023-08-18
include simg2dontcare image in manifest

adeebshihadeh:


internal/package_ota.py
```diff
@@ -0,0 +1,70 @@
...
+  return subprocess.check_output(["sha256sum", fn]).decode().split()[0]
+
+
+def process_file(remote_url, fn, name, sparse=False, full_check=True, has_ab=True, alt=None) -> dict:
+  hash_raw = hash = checksum(fn)
+  size = os.path.getsize(fn)
+
+  if sparse:
+    with NamedTemporaryFile() as f:
+      os.system(f"simg2img {fn} {f.name}")
```
adeebshihadeh, line 27: use subprocess, this doesn’t check return code

internal/package_ota.py
```diff
@@ -0,0 +1,70 @@
...
+      hash_raw = checksum(f.name)
+      size = os.path.getsize(f.name)
+
+  # TODO: compress
```
adeebshihadeh, line 31: needs a bit of a refactor so we don’t compress twice for production and staging

internal/package_ota.py
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env python3
+import json
+import os
+import subprocess
+from copy import deepcopy
+from pathlib import Path
+from tempfile import NamedTemporaryFile
+from typing import Optional, Union
```
adeebshihadeh, line 8: no need to type a tiny self-contained script like this, just adds bloat

adeebshihadeh:
LGTM

PR #180, opened 2023-11-07
hostname with serial number

gregjhogan:
This even allows using `hostnamectl` to change the hostname (although it changes back to serial number on reboot)
```sh
comma@tici-d62c206c:/data/openpilot$ hostnamectl 
   Static hostname: tici-d62c206c
         Icon name: computer
        Machine ID: 0e56b517e1dbc6ffa9481f7263f690e5
           Boot ID: b86719b5f5904aeeab49a8e89ea8abdf
  Operating System: Ubuntu 20.04.5 LTS
            Kernel: Linux 4.9.103+
      Architecture: arm64

comma@tici-d62c206c:/data/openpilot$ nmcli general hostname
tici-d62c206c
```

Note that the following alternative approaches do not work:
* `hostnamectl set-hostname --transient <new-host-name>` because `NetworkManager` doesn't respect a pre-existing transient hostname.
* `nmcli general hostname <new-host-name>` because it seems impossible to hook it after `NetworkManager` is up and before it brings up network interfaces

(both approaches set the hostname, but not in a way that the hostname reliably gets passed through DHCP for registration in DNS)

TODO:
- [x] test builder change to add symlink
- [x] change from `tici-` to `comma-`

### DIFF
--- a/Dockerfile.agnos
+++ b/Dockerfile.agnos
@@ -160,6 +160,9 @@ RUN cd /tmp/ && \
 COPY ./userspace/files/linux-headers-4.9.103+_4.9.103+-1_arm64.deb /tmp/
 RUN dpkg -i /tmp/linux-headers-4.9.103+_4.9.103+-1_arm64.deb
 
+# use hostname from kernel everywhere (nomrally transient)
+RUN rm /etc/hostname && ln -s /proc/sys/kernel/hostname /etc/hostname
+
 # Weston with hacked touch rotate and color correction
 COPY ./userspace/files/weston /usr/bin/weston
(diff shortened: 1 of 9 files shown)


gregjhogan:
note that if we ever get to kernel `5.19` there is a `hostname=` [kernel parameter](https://www.kernel.org/doc/html/latest/admin-guide/kernel-parameters.html#:~:text=BUGS%3DARM%2CSH%5D-,hostname,-%3D%20%20%20%20%20%20%20%5BKNL%5D%20Set%20the)

gregjhogan:
If we want to ensure the hostname defaults to `tici` the only sane way I could find to achieve this was to put it in the kernel config:
https://github.com/commaai/agnos-kernel-sdm845/pull/58

PR #181, opened 2023-11-11
Userspace: sync pytest depends

jnewb1:


userspace/pyproject.toml
```diff
@@ -93,7 +97,4 @@ flaky = "*"
 [build-system]
 requires = ["poetry-core"]
 build-backend = "poetry.core.masonry.api"
-pytest-mock = "*"
```
jnewb1, line 96: These two (mock and repeat) weren't actually included in the build because they are in the dev dependencies section, and are not in the main repo anywhere

jnewb1: nvm, i see they are used in the panda tests

PR #182, opened 2023-11-18
ubuntu 23.10

adeebshihadeh:
prepping for the 20.04 -> 24.04 update

### DIFF
--- a/Dockerfile.agnos
+++ b/Dockerfile.agnos
@@ -2,7 +2,7 @@
 # ###### Base ###### #
 # ################## #
 FROM scratch AS agnos-base
-ADD ubuntu-base-20.04.1-base-arm64.tar.gz /
+ADD ubuntu-base-23.10-base-arm64.tar.gz /
 
 # Add aarch64 and arm support
 COPY --from=multiarch/qemu-user-static:x86_64-aarch64 /usr/bin/qemu-aarch64-static /usr/bin
--- a/build_system.sh
+++ b/build_system.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 set -e
 
-UBUNTU_BASE_URL="http://cdimage.ubuntu.com/ubuntu-base/releases/20.04/release"
-UBUNTU_FILE="ubuntu-base-20.04.1-base-arm64.tar.gz"
+UBUNTU_BASE_URL="http://cdimage.ubuntu.com/ubuntu-base/releases/23.10/release"
+UBUNTU_FILE="ubuntu-base-23.10-base-arm64.tar.gz"
 
 export DOCKER_BUILDKIT=1
 
--- a/userspace/base_setup.sh
+++ b/userspace/base_setup.sh
@@ -85,8 +85,6 @@ apt-get install --no-install-recommends -yq \
     network-manager \
     nvme-cli \
     openssl \
-    python-dev \
-    python-setuptools \
     smartmontools \
     speedtest-cli \
     ssh \
--- a/userspace/base_setup.sh
+++ b/userspace/base_setup.sh
@@ -8,6 +8,15 @@ HOST=comma
 touch /TICI
 touch /AGNOS
 
+# Add armhf as supported architecture
+dpkg --add-architecture armhf
+
+# Install packages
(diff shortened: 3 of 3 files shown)


andiradulescu:
Any help needed here?

adeebshihadeh:
This is just some casual prep for 24.04. I'd be happy to merge ready-to-go PRs to further this prep, but won't have any time for reviews.

andiradulescu:
I will take a look to see what else is needed and try to build the image and even flash it. I need 22.04 on my C3 to test something. So, I won’t submit a random PR just to waste your time.

If I submit a ready-to-go PR on this, how can I/you validate it without needing a review?

adeebshihadeh:
I would approach it like this:
1. get >=22.04 Ubuntu working on a branch
2. open a PR for that just so we can see all the changes necessary
3. split out changes that we can start shipping now into their own PRs

#3 might or might now make sense depending on the changes. If nothing applies to 20.04 or makes sense to merge early, it'll still be valuable to have your PR up as a reference when I go to port too 24.04. To be clear, we're not gonna update the base version until 24.04 is out.

andiradulescu:
Agree. Started working on this here https://github.com/andiradulescu/agnos-builder/tree/2204 will get back to you when I have a working image or if I hit something (and/or stop).

andiradulescu:
I removed the bionic deps, since I was getting this error `GPG error: http://ports.ubuntu.com/ubuntu-ports bionic InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 3B4FE6ACC0B21F32` and updated the deps from [here](https://github.com/commaai/agnos-builder/pull/183#issuecomment-1848929094.) to the appropiate 22.04 versions.

Now the next hop is that `qt5-default` has been removed on 22.04: https://www.mail-archive.com/ubuntu-bugs@lists.ubuntu.com/msg5902890.html / https://askubuntu.com/a/1337334/562170

There are two options going forward:
- make/build qt5 on 22.04
- upgrade to qt6 (although I think we'll have some blockers e.g. `mapbox-gl-native-qt`)

I suggest making qt5 work (https://askubuntu.com/questions/1404263/how-do-you-install-qt-on-ubuntu22-04) for this PR and upgrade to qt6 in a separate future PR.

I will continue with qt5. Let me know what you think.

adeebshihadeh:
I think we have to stick with Qt5. Last time I looked into it, Qt6 didn’t have feature parity yet. If we build it ourselves, this is potentially something that can be merged early.

adeebshihadeh:
@andiradulescu how's your 22.04 port going? I see Ubuntu has started pushing 24.04 builds. If you're interested in switching to that, I can put up a bounty to do 24.04.

http://www.cdimage.ubuntu.com/daily-live/current/

andiradulescu:
@adeebshihadeh a bounty would help. I stopped when I ran out of free minutes (and space) on GitHub Codespaces and then I focused on something else.

If you put a bounty for 24.04, I will come back to this asap.

adeebshihadeh:
Hmm, actually doesn't seem like they're pushing the ARM base images we need for 24.04, unless I'm missing something. I think the best we can do is 23.10 right now.

Let's do $500 for an AGNOS based on 23.10:
* full openpilot builds
* passes openpilot's `selfdrive/test/test_onroad.py`
* everything works except 1-2 very specific things (e.g. modem and audio output)

I think it'll be hard to get everything working reliably, which is why it's $500 and ok if a couple things don't work yet.

andiradulescu:
@adeebshihadeh sounds good! I can resume immediately.

As for ARM base image I think this is the base image: https://cdimage.ubuntu.com/ubuntu-base/daily/current/**noble-base-arm64.tar.gz**

Is the bounty considered locked? If I fail or get blocked I will let you know.

adeebshihadeh:
Oh cool, let's use that 24.04 image then. And yes, it's locked to you. Can you put up a PR so I can follow along :)

andiradulescu:
https://github.com/commaai/agnos-builder/pull/188 thanks! ;)

adeebshihadeh:
Closing this then, good luck!

PR #188, opened 2024-01-15
Ubuntu 24.04

andiradulescu:
@adeebshihadeh starting 24.04 from the previous [22.04](https://github.com/andiradulescu/agnos-builder/tree/2204) branch.

PR #183, opened 2023-11-18
remove bionic deps

adeebshihadeh:

PR #191, opened 2024-01-19
Remove bionic deps

andiradulescu:
As written [here](https://github.com/commaai/agnos-builder/pull/183#issuecomment-1848929094) and [here](https://github.com/commaai/agnos-builder/pull/188#issuecomment-1899033458), the bionic deps removal implies updating a couple of `armhf` deps and adding `armhf` [libffi.so.6.0.4](https://launchpad.net/ubuntu/bionic/armhf/libffi6/3.2.1-8) from bionic. `libffi6` is required by [weston](https://github.com/commaai/agnos-builder/pull/188#issuecomment-1899022512).


adeebshihadeh:
Thanks!

andiradulescu:
@adeebshihadeh you are welcome! :) just that we have a "small" hiccup - I think ssh server is not up in this image (or it crashes), I can't ssh into it:

`ssh: connect to host 192.168.1.XXX port 22: Connection refused`

It boots, it has wifi, seems to work "fine", besides ssh. I'm trying to connect via usb-serial, but, at least on mac, it doesn't work.

I will try to figure it out tomorrow and submit a new PR.

adeebshihadeh:
Noticed the same thing, but was able to start it with `systemctl start ssh`

andiradulescu:
Reverted https://github.com/commaai/agnos-builder/pull/164 manually and now I have `ssh` on startup.

This after undoing my changes manually without success. I am currently building a clean image from https://github.com/commaai/agnos-builder/pull/164 to be 100% sure.

Though, for some reason, in my 24.04 image with https://github.com/commaai/agnos-builder/pull/164 ssh works on startup. Can't understand why and building images is slow to test all possible suppositions.

PR #194, opened 2024-01-20
Start ssh early

andiradulescu:
My proposal for fixing ssh on boot. Building another image to test.


andiradulescu:
@adeebshihadeh saw that you reverted https://github.com/commaai/agnos-builder/commit/a5b2f2b75ef0dae948f32469f0d355d140819780. we can merge this instead for starting ssh early. I tested it and seems fine. I can make some checks how early it starts now, if you want.

adeebshihadeh:
Thanks!

PR #201, opened 2024-01-24
Fastboot -> QDL flashing

bongbui321:
Transferring scripts to EDL. 
Also closes [#217](https://github.com/commaai/agnos-builder/issues/217)


install_edl.sh
```diff
@@ -0,0 +1,10 @@
+#! /bin/bash
```
adeebshihadeh, line 1: this didn't work on my machine
```
batman:agnos-builder$ ./install_edl.sh
Cloning into 'edl'...
Warning: Permanently added 'github.com' (ECDSA) to the list of known hosts.
remote: Enumerating objects: 2353, done.
remote: Counting objects: 100% (2353/2353), done.
remote: Compressing objects: 100% (775/775), done.
remote: Total 2353 (delta 1556), reused 2178 (delta 1540), pack-reused 0
Receiving objects: 100% (2353/2353), 63.09 MiB | 14.77 MiB/s, done.
Resolving deltas: 100% (1556/1556), done.
Submodule 'Loaders' (https://github.com/bkerler/Loaders) registered for path 'Loaders'
Cloning into '/home/batman/agnos-builder/edl/Loaders'...



Warning: Permanently added 'github.com' (ECDSA) to the list of known hosts.
Warning: Permanently added 'github.com' (ECDSA) to the list of known hosts.
remote: Total 0 (delta 0), reused 0 (delta 0), pack-reused 0
Warning: Permanently added 'github.com' (ECDSA) to the list of known hosts.
remote: Enumerating objects: 21, done.
remote: Counting objects: 100% (21/21), done.
remote: Compressing objects: 100% (12/12), done.
remote: Total 12 (delta 7), reused 5 (delta 0), pack-reused 0
Unpacking objects: 100% (12/12), 623.47 KiB | 1.57 MiB/s, done.
From github.com:bkerler/Loaders
 * branch            a83850577f32faaf4694cda8f00666e0d9d34654 -> FETCH_HEAD
Submodule path 'Loaders': checked out 'a83850577f32faaf4694cda8f00666e0d9d34654'
/home/batman/.pyenv/versions/3.11.4/bin/python: No module named pip3
```

flash_all.sh
```diff
@@ -4,13 +4,18 @@ set -e
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null && pwd)"
 cd $DIR
 
+EDL_DIR=$DIR/edl
```
adeebshihadeh, line 7: let's remove this from the flashing scripts. just like fastboot, you can assume the user ran the setup script

flash_all.sh
```diff
@@ -4,13 +4,18 @@ set -e
...
+  $EDL_DIR/edl w ${part}_a $DIR/agnos-firmware/$part.bin
+  $EDL_DIR/edl w ${part}_b $DIR/agnos-firmware/$part.bin
 done
 
 ./flash_bootloader.sh
 ./flash_kernel.sh
 ./flash_system.sh
 
-fastboot continue
+$EDL_DIR/edl reset
```
adeebshihadeh, line 21: i'd make the env var directly point to `edl` like this
```suggestion
$EDL reset
```

install_edl.sh
```diff
@@ -0,0 +1,10 @@
...
+
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null && pwd)"
+EDL_DIR=$DIR/edl
+
```
adeebshihadeh, line 5: the setup should also be part of the normal extract_tools script so you only have to run one thing to get setup

bongbui321:
Updated it. If possible, can you test it on the Comma 3/3x? I'm a bit concerned about the erase command of userdata since it has been known to be brutally [slow](https://github.com/bkerler/edl/issues/402#issuecomment-1879994867). From what I read from the code, it sends XML file as commands through firehose and write NULL to the partition. I would be careful about siginting the process if it is too slow. if I sigint it during the edl process then I had to restart it manually

andiradulescu:
C3/C3X uses UFS, but you don't need to specify it manually, since it autodetects it (if you don't specify `--memory=`).

I can't wipe my C3 right now, to test the erase. But, if the erase is slow on your device, most likely it will be the same on C3/C3X.

How to quick erase it would need more research. I think the best way is to ask the [creator]([bkerler](https://github.com/bkerler)) of https://github.com/bkerler/edl. You can open a [new issue](https://github.com/bkerler/edl/issues/new) or comment on an existing issue. I would open a new issue.

bongbui321:
@andiradulescu yeah but there were issues when I was flashing the phone when it auto detects emmc, not sure why so i would just safe guard it by declaring it

adeebshihadeh:
The erase and "skip chunks" image is just an optimization for how fastboot is implemented in our ABL. The erase can just be skipped if you use the full system image.

bongbui321:
From my understanding, `userdata` and `cache` partition is different from `system` partition, if we want a factory reset then we would have to erase both the `userdata & cache` partitions, which are separate from `system` partition no? If that is the case, then we would still need `erase` command

adeebshihadeh:
Correct, but those are much smaller. And preferably we format those partitions.

bongbui321:
yes, the thing that is superslow right now is the erase of the `userdata` partition. and I'm not sure how we can format the partition with the available command from the edl repo. From what I read from the `commaai/flash` repo, it uses `erase` for the userdata and cache partition as well

flash_all.sh
```diff
@@ -3,14 +3,20 @@ set -e
 
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null && pwd)"
 cd $DIR
+EDL=$DIR/edl/edl
+
+flash() {
```
adeebshihadeh, line 8: instead of copy/pasting this function in all the scripts, let's put it in a script called `tools/edl_flash.sh`

adeebshihadeh (commented):
Looks pretty good, but it'll look a lot cleaner if we put all the complexity of the edl running in a wrapper script in `tools/`

flash_all.sh
bongbui321: agree

adeebshihadeh:
Looks like `edl` still exits with 0 when it fails.

flash_all.sh
adeebshihadeh: this is more what i was thinking: https://github.com/commaai/agnos-builder/pull/201/commits/82abb831316f82b50ae61594ae0d68d69c30ed90

adeebshihadeh:
@bongbui321 can you fix edl not exiting with a bad return code on failure? then this is good to merge

If you're interested, I'll put up another $300 bounty to implement the fastboot CLI in edl. I don't think it'll require any new functionality; it should just be adding that new CLI api.

bongbui321:
Sorry, forgot to notify you abt that bug. I PRed a fix yesterday https://github.com/bkerler/edl/pull/531. This was from when you already in active slot a and try to set active slot again, previously this wasn't checked. Other EDL PRs shouldn't have this issue since it checks for activeslot before switching.

if needed you can use this command to get the correct gpt header images : `./edl gpt <holder_dir_for_gpt_header_images> --memory=ufs`

then flash the 6 luns correctly using those images: 
`./edl w gpt <holder_dir_for_gpt_header_images>/<correct_gpt_image_for_that_lun> --memory=ufs --lun=<lun_num>`

Sorry not on my computer rn, so don't have the script that I use.

bongbui321:
> If you're interested, I'll put up another $300 bounty to implement the fastboot CLI in edl. I don't think it'll require any new functionality; it should just be adding that new CLI api.

@adeebshihadeh sure.

adeebshihadeh:
> > If you're interested, I'll put up another $300 bounty to implement the fastboot CLI in edl. I don't think it'll require any new functionality; it should just be adding that new CLI api.
> 
> @adeebshihadeh sure. I would try to get all the current edl bounties merged first.

Sounds good https://github.com/commaai/agnos-builder/issues/217

adeebshihadeh:
Nice job upstreaming all these fixes!

bongbui321:
I'll continue finishing to make it less verbose when I have free time a bit busy these coming weeks (already made a PR but still isn't in a state I want).

PR #203, opened 2024-01-25
bring factory reset back to userspace

adeebshihadeh:
prereq to removing the fastboot timeout in the ABL


devtekve:
@adeebshihadeh this introduced a behavior where calling `sudo systemctl restart comma` would always bring up the system reset dialog. However, the reset dialog does not ask the screen to turn on if the screen is already off. Leading to a full black screen in those cases, and leading to an annoying reset prompt otherwise, because the count of taps is from device boot and never resets.

Issue #16, opened 2021-11-15
[$5000 Bounty] Use Upstream Weston


VirtuallyChris:
We currently use the closed source Thundercomm version of Weston.

Get open source Weston working in AGNOS for a $5000 bounty.
https://github.com/wayland-project/weston

geohot:
Has to run openpilot as normal to qualify

robbederks:
It actually turns out all the code we use (except for some minor internal changes) is in fact open source, albeit horribly outdated.
For reference, these are the public remotes used to build the weston we use now:
- https://source.codeaurora.org/platform/hardware/qcom/display: `LE.UM.2.3.7`
- https://source.codeaurora.org/platform/external/libdrm: `aosp-new/nougat-release`
- https://source.codeaurora.org/quic/le/wayland/weston: `LE.UM.2.3.7`

kfatehi:
Hi. I took a shot at this out of curiosity and failed, but I got some potentially useful output and scripts out of it, having been able to execute a custom built weston 9 on the tici at the end of it. Documented a post-mortem of my attempt here with the relevant error logs: https://github.com/kfatehi/agnos-builder/blob/3a3e0839bbba9c6d1015915e8f999a57e990fdd0/weston/README.md

houqp:
@robbederks is the corresponding mesa freedreno gallium driver source also open-sourced?

robbederks:
> @robbederks is the corresponding mesa freedreno gallium driver source also open-sourced?

We aren't using freedreno, but rather the proprietary Adreno shared libraries (see the content of the display deb package). Making this part open source is obviously not in the scope of this bounty.

houqp:
OK, turns out unlike the open-source mesa gbm, the proprietary version has the gbm driver backend statically linked into the provided `libgbm.so` shared library. @kfatehi your setup linked weston with ubuntu provided gbm lib, which loads driver backend from `lib/dri/*_dri.so` files. This is why you are were getting that `MESA-LOADER: failed to open msm_drm` error.

@robbederks is the proprietary `gbm.h` header file available for share to help with the linking?

robbederks:
[...]

PR #205, opened 2024-01-30
Update Weston

andiradulescu:
@adeebshihadeh this is my proposal for https://github.com/commaai/agnos-builder/issues/16

This will update `weston` to the Linaro latest downstream (based on weston 8.0.0) for Ubuntu 20.04, used in Qualcomm Robotics RB5 Dev Kit / QRB5165. The manifest (and tag) used is [LU.UM.3.3.1.r1-25900-QRB5165.0](https://git.codelinaro.org/clo/le/wayland/weston/-/commits/LU.UM.3.3.1.r1-25900-QRB5165.0?ref_type=tags).

My proposal comes for a couple of reasons:
- using upstream weston (as it is) forces to use mesa/freedreno which has [sketchy](https://mesamatrix.net/#RusticlOpenCL) support for OpenCL
- using upstream weston with the adreno drivers would mean to put (the linaro) [patches](https://git.codelinaro.org/clo/le/wayland/weston/-/compare/8.0.0...LU.UM.3.3.1.r1-25900-QRB5165.0?from_project_id=22479&straight=false) on top of the upstream weston - which is the same thing as pulling the linaro downstream - if we want this, we could probably reduce the number of changes Linaro did (e.g. only import the [sdm_backend](https://git.codelinaro.org/clo/le/wayland/weston/-/commit/ee1cba18e6bdb04e48bbb7eacbb6f4a57994ec8a)), but there will be a lot of trial and error
- using weston 8 since Ubuntu 20.04 [uses](https://packages.ubuntu.com/focal/weston) weston 8

Please let me know what you think.


adeebshihadeh:
Bounty is locked! I'll have to take some time to review the approach and make sure it hits all the things we want to do.

andiradulescu:
Thanks! Sure, take your time. I will putting this on hold for at least a couple of days.

Current progress is that all `compile-*.sh` steps compile fine, except `weston` which has some linking errors that need to be fixed next:
```
12.11 /usr/bin/ld: warning: liblog.so.0, needed by compositor/libexec_weston.so.0.0.0, not found (try using -rpath or -rpath-link)
12.11 /usr/bin/ld: compositor/libexec_weston.so.0.0.0: undefined reference to `__android_log_print'
12.11 /usr/bin/ld: libweston/libweston-8.so.0.0.0: undefined reference to `gbm_perform'
```

After `weston` compiles fine, I need to push the new version(s) of weston to my C3 and fix every problem one by one.

At the end, any private changes you have on the current `weston`, need to be copied over (as patches) to this PR.

andiradulescu:
If you need it, the entire `LU.UM.3.3.1.r1-25900-QRB5165.0` Linaro release can be pulled with:
```
repo init --depth=1 --current-branch -u https://git.codelinaro.org/clo/le/le/manifest.git -b release -m LU.UM.3.3.1.r1-25900-QRB5165.0.xml --repo-url=https://git.codelinaro.org/clo/tools/repo.git --repo-branch=qc-stable
repo sync -j8 -c
```
This is from where I extracted manually all the compile phases. By hand.

andiradulescu:
Last commits pack `weston` (and deps) in deb files, with applied patches.

Weston now loads fine, UI works except there is a lot of tearing because `backend-sdm` expects the kernel to support `atomic_modeset` for `drm`/`msm`/`adreno` drivers. From what I researched, it seems the support needed is from `4.19`. Our kernel is based on `4.9`.

What I'm currently trying to do, is finding/making a "fix" so that `backend-sdm` works without `atomic_modeset`. Since in weston 1.9.0 `compositor-sdm.c` doesn't need this.

Keep in mind that if you build from this branch, weston debs are not installed automatically (they are placed in `/tmp`, right now). I'm leaving this for the end.

adeebshihadeh:
> * using weston 8 since Ubuntu 20.04 [uses](https://packages.ubuntu.com/focal/weston) weston 8

Which version does 24.04 use? Should we just go straight to that version?

andiradulescu:
Ubuntu [24.04](https://packages.ubuntu.com/noble/weston) uses the latest Weston [13.0.0](https://wayland.freedesktop.org/releases.html). We can probably go straight to the latest.

The only issue right now is with the kernel drm to support atomic modesetting:
- either patch the current kernel (putting a newer drm on top)
- either update the kernel (to a version where it's easy/fast to apply all [changes](https://github.com/commaai/agnos-kernel-sdm845/compare/3666be0..c22355e) as patches - e.g. msm-4.19 or msm-5.4 from Linaro)
- find a fix (e.g. downgrade the `backend-sdm` to a version without atomic modesetting)

The next thing I'm trying is to patch the current kernel.

Mainlining the kernel will probably take more time than using a Linaro "downstream" msm kernel. But, there is some effort here https://github.com/robin-reckmann/comma3_mainline / https://github.com/commaai/agnos-builder/issues/16#issuecomment-1068000348 that needs to be tested/compared, since I have no idea if the "msm" specific stuff we need is already in mainline (and what isn't). My effort is towards using the sdm/drm adreno drivers (not freedreno).

adeebshihadeh:
I'm hesitant to put more effort into the current kernel, since we'd like to go to mainline soon-ish. For mainline, the very latest or the 24.04 kernel are probably what we'd want to go with.

andiradulescu:
> I'm hesitant to put more effort into the current kernel, since we'd like to go to mainline soon-ish. For mainline, the very latest or the 24.04 kernel are probably what we'd want to go with.

Just so you know, I am/was very hesitant in doing any work to make things work on anything except Ubuntu 24.04 / latest.

Indeed, I tried some easy fixes to see weston work 100% but spent little time actually doing this.

As for this branch, it has very little backward compatibility fixes (just some missing include files in kernel 4.9). They can be easily deleted. You can check the diff.

Next, I would see the following get done:
- clean/update old dependencies (including removal of armhf deps)
- mainline the kernel
- update weston 8 to weston 13

And this should be done, or close to done, in the Ubuntu 24.04 release window aka until April 25, 2024.

andiradulescu:
The last conclusion of this PR on “old” 4.9 kernel is that it works but with lots of tearing, since there is no atomic modesetting which this version of weston with backend-sdm needs.

andiradulescu:
Closing this PR, since changes in this PR, as they are, should not be needed anymore in mainline. More will be seen when weston is getting some work in mainline.

PR #215, opened 2024-04-12
Update README.md

eatfishfish:
build docker image.


adeebshihadeh:
`./build_system.sh` is designed to do this.

PR #221, opened 2024-04-30
Ubuntu 24.04

adeebshihadeh:
Try it out with this openpilot branch: https://github.com/commaai/openpilot/pull/32330


adeebshihadeh:
@andiradulescu will maintain the canonical 24.04 PR

PR #231, opened 2024-06-19
Build kernel in CI

### LINKED ISSUE #230: [$250 bounty] System image and kernel building in CI
Ideally this runs on an arm64 runner - not sure if GHA offers those on the open source tier though.

Requirements
* `build_system.sh` and `build_kernel.sh` running in CI
* uploads the built images as an [artifact](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts)
* must have incremental caching between runs so the avg run is (somewhat) fast

Bounty is in two parts:
- [x] kernel build in CI ($100) https://github.com/commaai/agnos-builder/pull/231
- [x] system build in CI ($150)

incognitojam:
- [x] build kernel
  - [x] use ccache (seems to reduce from 13 -> 3 mins)
  - [x] upload `boot.img` artifact
- [x] build system
  - [x] switch to arm64 custom runner
  - [x] pull/cache docker steps (need to wait for successful run?)

closes #230


adeebshihadeh:
The private submodules aren't needed to run the kernel and system build.

incognitojam:
Oh my bad, was reading package_ota

incognitojam:
There is a beta for orgs on GitHub team plan and above to use arm runners, but I think it has to be configured in the **organisation settings** -> actions -> runners -> New GitHub-hosted runner

Instructions: https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners
Announcement: https://github.blog/changelog/2024-06-03-actions-arm-based-linux-and-windows-runners-are-now-in-public-beta/

adeebshihadeh:
Even with the free GHA arm64 runners, it seems like we're not gonna get that working in a reasonable amount of time.

We use https://namespace.so/ runners for some openpilot CI. I'm ok using a big arm64 machine in namespace if it'll work for external PRs too.

andiradulescu:
@incognitojam not looking to steal your bounty, just giving a hand - take a look over these changes - https://github.com/andi-radulescu/agnos-builder/compare/3edc543..72998211

I have a build running that should not fail anymore - https://github.com/andi-radulescu/agnos-builder/actions/runs/9651664681

Between stuff today I played with https://namespace.so - created a profile that looks like this:

image

The arm64 4x16 machines builds the kernel and system (without caching) in a little over an hour.

incognitojam:
Cool, I hadn't tried creating a custom profile myself

I don't understand why github actions gets stuck on `Waiting for a runner to pick up this job...`

adeebshihadeh:
> Cool, I hadn't tried creating a custom profile myself
> 
> ~I don't understand why github actions gets stuck on `Waiting for a runner to pick up this job...`~ actually it might need comma to link the repo to namespace.so

@maxime-desroches can get this setup?

incognitojam:
Runner Profile:
- linux/arm64
- 8x16 (at least)
- Ubuntu 20.04
- Enable all caching toggles (container images, git repo checkouts, toolchain downloads)

maxime-desroches:
@incognitojam should be good to use `namespace-profile-arm64-8x16-2004-caching`

andiradulescu:
Some insight about what I did [yesterday](https://github.com/andi-radulescu/agnos-builder/commits/ci/):
- tools is not needed on native builds
- build_kernel.sh needs python2 and 20.04 (because python2/libssl errors on 22.04)
- the namespace runners have some permission issues (even when using sudo - can't delete /opt/hostedtoolcache; can't use mount/unmount
- added mounting inside a(nother) docker container (because of the above issue)
- lfs files weren't pulled (e.g. *.deb files for system)
- added missing deps (bc and img2simg)

Added 2 more commits:
- tar was failing on yesterday's last build (some tar bug) - so I moved untaring inside the docker container - still waiting for the build to finish
- build_system.sh now "almost" works natively on Apple M2 with Docker/OrbStack (without running in a VM) - I'll leave the full support for a separate PR

adeebshihadeh:
I think it should be reliable after #234

adeebshihadeh:
`pyopencl` still had some issue, so I just disabled it on master for now.

incognitojam:
It finally works!
image

incognitojam:
Should be able to get it ~5 mins quicker by caching the kernel build step

andiradulescu:
Glad to see this working @incognitojam!

A further improvement I would do, since only `system.img` is uploaded as an artifact, is to not build `system-skip-chunks.img` anymore.

The `alt` image creation I think it can safely be moved to `package_ota.py` and `flash_system.sh` can just use `system.img`, since both fastboot and edl can flash `system.img`.

What do you think?

build_system.sh
```diff
@@ -46,45 +48,50 @@ echo "Building image"
 export DOCKER_CLI_EXPERIMENTAL=enabled
 docker build -f Dockerfile.agnos -t agnos-builder $DIR
 
+# Setup mount container
+MOUNT_CONTAINER_ID=$(docker run -d --privileged --volume $BUILD_DIR:$BUILD_DIR ubuntu:latest sleep infinity)
```
adeebshihadeh, line 52: what's this needed for?

.github/workflows/build.yaml
```diff
@@ -0,0 +1,72 @@
...
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        lfs: true
+
+    - name: get kernel submodule ref
+      id: kernel-submodule
+      run: echo "ref=$(git ls-tree HEAD | awk '$4 == "agnos-kernel-sdm845"' | awk '{print $3}')" >> "$GITHUB_OUTPUT"
+
+    - uses: namespacelabs/nscloud-checkout-action@v4
```
adeebshihadeh, line 29: is this faster than the normal checkout action?

.github/workflows/build.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+    - name: Install dependencies
+      run: sudo apt-get update && sudo apt-get install -y bc img2simg
+
+    - name: Install python2
+      uses: gabrielfalcao/pyenv-action@2f49ca7587f9d0663d13f1147b78d3361417eaf7
+      with:
+        default: '2.7.18'
+
+    - name: Cache kernel build artifacts
```
adeebshihadeh, line 43: we want incremental caching, otherwise it's a full rebuild for any development work

build_kernel.sh
```diff
@@ -16,13 +17,15 @@ if git submodule status --cached agnos-kernel-sdm845/ | grep "^-"; then
...
+if [ "$ARCH" != "arm64" ] && [ "$ARCH" != "aarch64" ]; then
+  $DIR/tools/extract_tools.sh
 
-# Build parameters
-export ARCH=arm64
-export CROSS_COMPILE=$TOOLS/aarch64-linux-gnu-gcc/bin/aarch64-linux-gnu-
-export CC=$TOOLS/aarch64-linux-gnu-gcc/bin/aarch64-linux-gnu-gcc
-export LD=$TOOLS/aarch64-linux-gnu-gcc/bin/aarch64-linux-gnu-ld.bfd
+  # Build parameters
+  export ARCH=arm64
```
adeebshihadeh, line 24: ARCH should still be set since it's read by the kernel makefiles

build_system.sh
andiradulescu: the namespace runners have some permission issues, even when using sudo - can't use (sudo) mount/unmount - so you mount them in a container (also works on mac with this change)

build_kernel.sh
andiradulescu: when ARCH is not set, it will build on the architecture of the system doing the build. on anything else (except arm64/aarch64), ARCH will be exported

.github/workflows/build.yaml
incognitojam: only 5s faster

adeebshihadeh:
So the kernel is building in CI and is somewhat fast without incremental caching. Let's split up the bounty and get that merged first? $100 for kernel build + $150 for the system build with incremental caching for both system/kernel

adeebshihadeh:
I split the bounty: #230, and now that a week has gone by, the other half of the bounty is unlocked for anyone who wants to pick it up.

andiradulescu:
@incognitojam do you still want to take the incremental caching?

incognitojam:
No you go ahead, sorry I didn't respond over the weekend, but I'm happy it was split anyway :)

Issue #5, opened 2021-09-28
Don't ship sparsified OTA images


adeebshihadeh:
This is just for setting sparse=False for the system image openpilot downloads in agnos.json

andiradulescu:
Will take care of this as soon as I finish with https://github.com/commaai/agnos-builder/pull/246

andiradulescu:
I'm a little confused on which type of image:
- raw
- sparsified
- sparsified with skip chunks

flashes in how much time with:
- fastboot
- edl

I will use the same host, cable and device, run some comparisons and come up with a table.

andiradulescu:
Can you shortly explain the relation to https://github.com/commaai/openpilot/issues/31494? Or the compatibility between the two?

What will `agnos-builder` ship when both issues are `Closed`?

PR #246, opened 2024-07-16
Improve build_system.sh mounting and caching

andiradulescu:


build_system.sh
```diff
@@ -86,22 +113,18 @@ DATETIME=$(date '+%Y-%m-%dT%H:%M:%S')
...
-docker rm -f $MOUNT_CONTAINER_ID > /dev/null
-
-cd $DIR
-
 # Sparsify
-echo "Sparsify image"
-TMP_SPARSE="$(mktemp)"
-img2simg $ROOTFS_IMAGE $TMP_SPARSE
-mv $TMP_SPARSE $SPARSE_IMAGE
+echo "Sparsify image $(basename $SPARSE_IMAGE)"
```
adeebshihadeh, line 117: I think we can just remove this https://github.com/commaai/agnos-builder/issues/5. want to make a separate PR to remove it and confirm it won't break anything?

IIRC we kept it around since the very first openpilot versions for C3 only worked with sparse images

andiradulescu: I wanted to suggest this also. I think `agnos-builder` (fastboot and edl), `comma flash` (fastboot and edl) and `openpilot updater` will work fine with the normal image. I don't see why not, we need to test.

When you say remove https://github.com/commaai/agnos-builder/issues/5 you mean remove both sparsified images, right?

andiradulescu: Actually, `system.img.raw` is 10GB, sp we need one sparsified image. I think we can just remove `system-skip-chunks.img`, since fastboot and edl supports both types of sparsified images and openpilot updater I think it uses the non-"skip-chunks" one.

andiradulescu:
Build is succesful on CI (with incremental caching) and also works on macOS.

I still need to be sure it still builds in usual linux/ubuntu arm64/x64.

Anything else I should change?

andiradulescu:
Building kernel is skipped completely if a previous build from the same `agnos-kernel-sdm845` commit ref has been built before. Since boot.img and kernel modules are cached.

adeebshihadeh (commented):
This is a really big PR. Can you split this up into smaller self-contained PRs?

andiradulescu:
Will split them, it's too big for my liking too. But it was the only way I could move fast.

Good idea with [upload] let's try it out.

build_system.sh
bongbui321: raw agnos is ~10GB so I think we still want sparse image for faster download. skip-chunk-img will speed up flash speed for fastboot since it would just skip and go to the next flash location. There is no difference in speed for QDL or even speed it up, QDL populates the data for the skip chunk area, removing it might saves some flashing time

.github/workflows/build.yaml
```diff
@@ -15,56 +15,82 @@ permissions:
...
+      id: cache-output
+      uses: actions/cache/restore@v4
+      with:
+        key: output-${{ steps.kernel-submodule.outputs.ref }}
+        path: |
+          output/boot.img
+          output/*.ko
+
     - name: Checkout agnos-kernel-sdm845
+      if: steps.cache-output.outputs.cache-hit != 'true'
```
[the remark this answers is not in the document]
andiradulescu: not sure what you mean. since `output/boot.img` and `output/*.ko` has just been restored from cache for the exact kernel ref, checking out the agnos-kernel-sdm845 submodule is for nothing since it will not be re-built.

andiradulescu:
Started to split this, will close when all other PRs are created.

andiradulescu:
New PRs:
- https://github.com/commaai/agnos-builder/pull/247
- https://github.com/commaai/agnos-builder/pull/249
- https://github.com/commaai/agnos-builder/pull/251

build_system.sh
andiradulescu: let's continue this discussion in https://github.com/commaai/agnos-builder/issues/5 or https://github.com/commaai/agnos-builder/pull/275

Issue #237, opened 2024-07-06
Support building AGNOS on macOS


adeebshihadeh:
Get `./build_kernel.sh` and `./build_system.sh` working on macOS

(This is for M1, not Intel macs)

andiradulescu:
https://github.com/commaai/agnos-builder/pull/246 also takes care of this, as soon as it's ready for review

PR #247, opened 2024-07-17
Build kernel in docker

andiradulescu:
For https://github.com/commaai/agnos-builder/issues/237:
- [x] builds on M-series Macs
- [x] builds on Intel Macs
- [x] builds in CI
- [x] build in Ubuntu 24.04 (aarch64)
- [x] README on how to use/create a `Case-sensitive APFS` volume on macOS

|Kernel build|ccache empty|ccache from previous build|ccache + out|
| :---:        |    :----:   |          :---: |           :---: |
|CI|<7min|<2min| - |
|local aarch64|<15min|<3min|<25sec|


build_kernel.sh
```diff
@@ -10,62 +11,96 @@ OUTPUT_DIR=$DIR/output
 BOOT_IMG=./boot.img
 cd $DIR
 
+if [[ "$(uname)" == 'Darwin' ]]; then
```
adeebshihadeh, line 14: any way we can work around this? we're already building in a docker?

andiradulescu: the only way to go around the "Case-sensitive APFS volume" (just to build) is to clone the kernel inside the container build it there and just copy the output/boot.img and *.ko files. or make just "output" a volume.

the problem is that if you want to make changes, you don't have seamless visibility from macOS on the files inside the container. the `agnos-builder/agnos-kernel-sdm845` will still be empty (not cloned). and you need to work in a container or VM.

if you want to work on the kernel, like any other project on your computer, it should stay on a "Case-sensitive APFS volume". this is how I work on the kernel. the "Case-sensitive APFS volume" is very nicely implemented in macOS.

Dockerfile.kernel
```diff
@@ -0,0 +1,29 @@
+FROM ubuntu:20.04
+
```
adeebshihadeh, line 2: can we just make one meta-builder Docker that also has the sparse stuff?

build_kernel.sh
```diff
@@ -10,62 +11,92 @@ OUTPUT_DIR=$DIR/output
...
+  echo "---------------   macOS support   ---------------"
+  echo "Ensure you are in a Case-sensitive APFS volume to build the AGNOS kernel."
+  echo "https://github.com/commaai/agnos-builder?tab=readme-ov-file#development---macos"
+  echo "-------------------------------------------------"
+  echo "Press any key to continue or CTRL+C to abort..."
+  read -n 1 -s
```
adeebshihadeh, line 20: can we detect instead of showing a prompt on every build?

tools/extract_tools.sh
```diff
@@ -27,23 +20,32 @@ git lfs ls-files | awk '{print $2}' | grep "-" &>/dev/null && {
...
 LINARO_GCC_TARBALL=$LINARO_GCC.tar.gz
 GOOGLE_GCC_4_9_TARBALL=$GOOGLE_GCC_4_9.tar.gz
 EDK2_LLVM_TARBALL=$EDK2_LLVM.tar.gz
 SEC_IMAGE_TARBALL=$SEC_IMAGE.tar.gz
 
-if [ ! -d $LINARO_GCC ]; then
-  tar -xzf $LINARO_GCC_TARBALL &>/dev/null
-fi
+ARCH=$(uname -m)
+if [ "$ARCH" != "arm64" ] && [ "$ARCH" != "aarch64" ]; then
```
adeebshihadeh, line 34: why not use the same toolchains on arm64?

andiradulescu: tried it, doesn't work on arm64. works fine on x64. will test more and if I find a way will PR it separately.

keep in mind that previously was the same thing, I just moved the code around so that `extract_tools.sh` is called on every platform to check for git lfs. this toolchain was never used on arm64/aarch64.

build_kernel.sh
andiradulescu: resolved!

adeebshihadeh:
This makes a fully cached build take 41s vs 9s on my workstation. Where's that extra time from? That's a pretty big difference.

andiradulescu:
time build_kernel.sh from this branch compared to the one in master, on Ubuntu 24.04 x64with Docker? I’ll compare and see where the difference is.

adeebshihadeh:
I'm not comparing with Docker, just running `./build_kernel.sh` on master and with this branch on the same machine.

andiradulescu:
`build_kernel.sh` in this branch uses Docker, the building happens in the meta builder container with file system mounted as a volume.

adeebshihadeh:
Is that why it takes 4x longer? Can we make it faster?

adeebshihadeh:
I don't think this is worth splitting it up into two different ways to build. It's nice that everyone's running the same stuff. Let's just make it fast as possible.

adeebshihadeh:
`./build_kernel_headers.sh` doesn't work on mac. Should we merge it into `./build_kernel.sh`?

Otherwise, tried on a mac and LGTM!

andiradulescu:
Nice catch, indeed the actual kernel build does more steps in Docker vs the native build. On my computer is 1m10s vs 20s. To solve this specific issue I need more time.

I didn't treat `build_kernel_headers.sh` since I'm using it a little differently for the mainline. And in the system build we use a prebuilt kernel headers archive.

For now it sounds good to merge them together.

adeebshihadeh:
I just ran build kernel headers on my workstation and it takes more than 1m. We can actually just leave that alone for now since we don't really update it often.

andiradulescu:
Just did the same and it takes too much time. I will take care of it in mainline.

build_kernel.sh
```diff
@@ -10,62 +11,95 @@ OUTPUT_DIR=$DIR/output
...
+fi
+
+# Setup kernel build container
+echo "Building agnos-meta-builder docker image"
+docker build -f Dockerfile.builder -t agnos-meta-builder $DIR \
+  --build-arg UNAME=$(id -nu) \
+  --build-arg UID=$(id -u) \
+  --build-arg GID=$(id -g)
+echo "Starting agnos-meta-builder container"
+CONTAINER_ID=$(docker run -d -v $DIR:$DIR -w $DIR agnos-meta-builder)
```
adeebshihadeh, line 32: might fix the CI caching issue?
```suggestion
CONTAINER_ID=$(docker run -d -v $DIR:/tmp/builder -w /tmp/builder agnos-meta-builder)
```

andiradulescu: this was not the issue. the path is absolute in the GH action - always `/home/runner/work/agnos-builder/agnos-builder/`.

caching the "out" directory is completely ignored by the kernel build and I can't find how it's detecting that. nevertheless, I couldn't find any advice on caching the `out` directory. only ccache (or similar) approaches are recommended.

after I found the previous issue we had, ccache works much better now, will add a table with the build times in CI and on my machine.

build_kernel.sh
```diff
@@ -10,62 +11,102 @@ OUTPUT_DIR=$DIR/output
...
+
+  # Build arm64 arch
+  export ARCH=arm64
+
+  # Set ccache dir
+  export CCACHE_DIR=$DIR/.ccache
+
+  # Avoid LINUX_COMPILE_HOST to change on every run thus invalidating cache
+  # https://patchwork.kernel.org/project/linux-kbuild/patch/1302015561-21047-8-git-send-email-mmarek@suse.cz/
+  export KBUILD_BUILD_HOST="docker"
```
andiradulescu, line 65: @adeebshihadeh the issue with increasing cached build time in docker was because LINUX_COMPILE_HOST had the container id (which is different everytime). setting KBUILD_BUILD_HOST="docker" makes LINUX_COMPILE_HOST="docker" (instead of the container id).

andiradulescu:
@adeebshihadeh please test the comparison again

> This makes a fully cached build take 41s vs 9s on my workstation. Where's that extra time from? That's a pretty big difference.

The issue was LINUX_COMPILE_HOST having a different name everytime in docker.

adeebshihadeh:
Getting 15s now, that's close enough

PR #249, opened 2024-07-18
Fix build system in docker on macOS

andiradulescu:
For https://github.com/commaai/agnos-builder/issues/237

Fixes some issue with macOS sparsify and some nice to haves.

### DIFF
--- a/build_system.sh
+++ b/build_system.sh
@@ -53,10 +53,8 @@ echo "Creating agnos-builder container"
 CONTAINER_ID=$(docker container create --entrypoint /bin/bash agnos-builder:latest)
 
 # Setup mount container for macOS and CI support (namespace.so)
-if ! docker inspect agnos-mount &>/dev/null; then
-  echo "Building agnos-mount docker image"
-  docker build -f Dockerfile.sparsify -t agnos-mount $DIR
-fi
+echo "Building agnos-mount docker image"
+docker build -f Dockerfile.sparsify -t agnos-mount $DIR
 echo "Starting agnos-mount container"
 MOUNT_CONTAINER_ID=$(docker run -d --privileged -v $DIR:$DIR agnos-mount)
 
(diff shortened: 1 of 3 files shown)


build_system.sh
```diff
@@ -38,38 +52,54 @@ if [ ! -f $UBUNTU_FILE ]; then
...
+
+# Create host user in container (fixes namespace.so error)
+USERNAME=$(whoami)
+docker exec $MOUNT_CONTAINER_ID bash -c "useradd --uid $(id -u) -U -m $USERNAME" &> /dev/null
 
 # Create filesystem ext4 image
 echo "Creating empty filesystem"
-docker exec $MOUNT_CONTAINER_ID fallocate -l $ROOTFS_IMAGE_SIZE $ROOTFS_IMAGE
-docker exec $MOUNT_CONTAINER_ID mkfs.ext4 $ROOTFS_IMAGE > /dev/null
+docker exec -u $USERNAME $MOUNT_CONTAINER_ID fallocate -l $ROOTFS_IMAGE_SIZE $ROOTFS_IMAGE
```
adeebshihadeh, line 85: i'd pull `docker exec -u $USERNAME $MOUNT_CONTAINER_ID` into a helper. it's super unreadable as is

adeebshihadeh (changes requested):

build_system.sh
```diff
@@ -38,70 +36,96 @@ if [ ! -f $UBUNTU_FILE ]; then
...
+  docker exec -u $USERNAME $MOUNT_CONTAINER_ID "$@"
+}
+
+exec_as_root() {
+  docker exec $MOUNT_CONTAINER_ID "$@"
+}
+
+# Create host user in container (fixes namespace.so error)
+USERNAME=$(whoami)
+exec_as_root useradd --uid $(id -u) -U -m $USERNAME &> /dev/null
```
robin-reckmann, line 72: This breaks when the host user has uid 1000 which already exists in the docker.

andiradulescu: good catch! since ubuntu 24.04 images have the ubuntu 1000:1000 user. I tested this with the kernel but not system, that’s it’s always good to have a checklist. thanks!

robin-reckmann: I have a fix and am current testing on both ubuntu and macos. PR should be up soon.

andiradulescu: @adeebshihadeh this needs to be merged https://github.com/commaai/agnos-builder/pull/252

PR #250, opened 2024-07-18
Fix python dependencies

robin-reckmann:
openpilot doesnt compile with the latest agnos master. The following changes make it work again.
- Re-enable pyopencl
- Downgrade numpy
- Add future-fstrings

### DIFF
--- a/userspace/pyproject.toml
+++ b/userspace/pyproject.toml
@@ -26,7 +26,7 @@ dependencies = [
   "libusb1",
   "markdown-it-py",
   "nose",
-  "numpy",
+  "numpy < 2.0.0",
   "onnx",
   "onnxruntime >= 1.16.3",
   "pillow",
@@ -39,7 +39,7 @@ dependencies = [
   "pydub",
   "pyjwt",
   "pylint",
-#"pyopencl == 2024.1", #FIXME: has some installation errors
+  "pyopencl == 2024.1",
   "pyserial",
   "pytest",
   "pytest-cov",
--- a/userspace/pyproject.toml
+++ b/userspace/pyproject.toml
@@ -17,7 +17,7 @@ dependencies = [
   "evdev",
   "flake8",
   "flask",
-  "future-fstrings",
+  "future-fstrings",  # needed for acados
   "gunicorn",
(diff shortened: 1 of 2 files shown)


userspace/pyproject.toml
```diff
@@ -39,7 +40,7 @@ dependencies = [
   "pydub",
   "pyjwt",
   "pylint",
-#"pyopencl == 2024.1", #FIXME: has some installation errors
+  "pyopencl == 2024.1",
```
adeebshihadeh, line 43: any idea why we have to pin? this is just a temp fix

userspace/pyproject.toml
```diff
@@ -26,7 +27,7 @@ dependencies = [
   "libusb1",
   "markdown-it-py",
   "nose",
-  "numpy",
+  "numpy < 2.0.0",
```
adeebshihadeh, line 30:
```suggestion
  "numpy < 2.0.0",  # pending bumping in openpilot
```

userspace/pyproject.toml
```diff
@@ -17,6 +17,7 @@ dependencies = [
   "evdev",
   "flake8",
   "flask",
+  "future-fstrings",
```
adeebshihadeh, line 20: what's this needed for?

robin-reckmann: acados

adeebshihadeh: can you leave a comment?

userspace/pyproject.toml
robin-reckmann: Any newer version also needs a newer version of cmake to compile. Cmake >= 3.17 is not available via apt in 20.04. We could pull it in via pip.

adeebshihadeh: that's fine, let's just leave a comment

robin-reckmann: You mean keep it pinned for now? We can probably unpin it with 24.04 later.

adeebshihadeh: yeah, but leave a comment about why it's pinned

PR #251, opened 2024-07-18
CI: add system build

andiradulescu:
For https://github.com/commaai/agnos-builder/issues/230:

- [x] increase timeout to 40min
- [x] upload artifacts only if [upload] tag is present in commit message
- [x] ccache action to reduce kernel build time
- [x] build_system.sh with docker layers cached in namespace
- [x] ubuntu-base image downloaded with `--remote-time` to persist docker layers

|Step|No cache|Cache|
| :---:        |    :----:   |          :---: |
|kernel|7min|3min*|
|system|33min|2min**|
|upload system image|4min|4min|

\* takes 3min with ccache - on my computer is much faster - caching the `out` folder doesn't make any difference - I assume this is because the absolute path where the kernel is built always changes between runs
\** depends a lot on what docker layers the changes have been made - if the change is on the first layers, it takes longer - if it's on the last layers, it takes less

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -27,15 +27,6 @@ jobs:
       id: kernel-submodule
       run: echo "ref=$(git ls-tree HEAD | awk '$4 == "agnos-kernel-sdm845"' | awk '{print $3}')" >> "$GITHUB_OUTPUT"
 
-    - name: Restore cache output boot.img and *.ko
-      id: cache-output
-      uses: actions/cache/restore@v4
-      with:
-        key: output-${{ steps.kernel-submodule.outputs.ref }}
-        path: |
-          output/boot.img
-          output/*.ko
-
     - name: Checkout agnos-kernel-sdm845
       uses: actions/checkout@v4
       with:
@@ -52,27 +43,22 @@ jobs:
         default: '2.7.18'
 
     - name: Restore cache kernel build
+      id: cache-kernel-out
       uses: actions/cache/restore@v4
       with:
-        key: kernel-out
+        key: kernel-out-${{ steps.kernel-submodule.outputs.ref }}
         path: agnos-kernel-sdm845/out
+        restore-keys: kernel-out-
 
     - name: Build kernel
       run: ./build_kernel.sh
 
     - name: Save cache kernel build
       uses: actions/cache/save@v4
+      if: always()
       with:
-        key: kernel-out
+        key: ${{ steps.cache-kernel-out.outputs.cache-primary-key }}
         path: agnos-kernel-sdm845/out
-
-    - name: Save cache output boot.img and *.ko
-      uses: actions/cache/save@v4
-      with:
-        key: output-${{ steps.kernel-submodule.outputs.ref }}
-        path: |
-          output/boot.img
-          output/*.ko
 
     - name: Upload artifact boot.img
       uses: actions/upload-artifact@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -15,7 +15,7 @@ permissions:
 
 jobs:
   build:
-    name: build kernel #and system
+    name: build kernel and system
     runs-on: namespace-profile-arm64-8x16-2004-caching
     timeout-minutes: 20
```
adeebshihadeh, line 20: is this timeout enough without a cache?

.github/workflows/build.yaml
```diff
@@ -42,29 +52,52 @@ jobs:
...
       with:
-        key: kernel-${{ steps.kernel-submodule.outputs.ref }}
-        path: |
-          agnos-kernel-sdm845/out
-        restore-keys: kernel-
+        key: kernel-out
+        path: agnos-kernel-sdm845/out
 
     - name: Build kernel
+      if: steps.cache-output.outputs.cache-hit != 'true'
```
adeebshihadeh, line 63: I don't like all these "skip if cache hit" since they add another case, and usually the time tradeoff isn't worth it.

build_system.sh
```diff
@@ -24,18 +24,29 @@ SKIP_CHUNKS_IMAGE="$OUTPUT_DIR/system-skip-chunks.img"
...
 
-# Copy kernel modules over
+# Check if kernel modules are built
 if ! ls $OUTPUT_DIR/*.ko >/dev/null 2>&1; then
-  echo "kernel modules missing. run ./build_kernel.sh first"
+  echo "Kernel modules missing. Run ./build_kernel.sh first"
   exit 1
 fi
+
+# Copy kernel modules over only if updated - otherwise prevents proper caching later
```
adeebshihadeh, line 33: when does this break the caching?

.github/workflows/build.yaml
andiradulescu: it should be ~40 minutes on namespace-profile-arm64-8x16-2004-caching. Updating to 40.

.github/workflows/build.yaml
andiradulescu: will remove them for now

build_system.sh
andiradulescu: in the last steps of the Dockerfile where `COPY ./userspace/usr/comma/ /usr/$USERNAME/` is copied then if the kernel module are always copied, then also the docker build stages run again (even if the modules are exactly the same)

andiradulescu: fyi, this adds very little time (seconds) to the build, but also creates additional layers that will accumulate to the namespace docker cache

adeebshihadeh: let's remove it. these kinds of micro-optimizations tend not to be worth it

.github/workflows/build.yaml
```diff
@@ -34,6 +48,13 @@ jobs:
         ref: ${{ steps.kernel-submodule.outputs.ref }}
         path: agnos-kernel-sdm845
 
+    - name: ccache
+      uses: hendrikmuhs/ccache-action@v1.2
```
adeebshihadeh, line 52: external actions should always be pinned to a commit
```suggestion
      uses: hendrikmuhs/ccache-action@c92f40bee50034e84c763e33b317c77adaa81c92
```

adeebshihadeh (approved):
Nice, thanks for posting the table too!

PR #252, opened 2024-07-19
Fix image creation docker workflow in host ubuntu

robin-reckmann:
useradd will fail if the host user has uid:gid 1000:1000. 
Fix: Delete user in container if the uid:gid is the same and only then create a new user with the same user name as the host user.

### DIFF
--- a/Dockerfile.sparsify
+++ b/Dockerfile.sparsify
@@ -1,9 +1,18 @@
 FROM ubuntu:24.04
 
+ARG UNAME
+ARG UID
+ARG GID
+
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && \
     apt-get install -y android-sdk-libsparse-utils && \
     rm -rf /var/lib/apt/lists/*
+
+RUN userdel -r `getent passwd ${UID} | cut -d : -f 1` > /dev/null 2>&1; \
+    groupdel -f `getent group ${GID} | cut -d : -f 1` > /dev/null 2>&1; \
+    groupadd -g ${GID} -o ${UNAME} && \
+    useradd -u $UID -g $GID ${UNAME}
 
 ENTRYPOINT ["tail", "-f", "/dev/null"]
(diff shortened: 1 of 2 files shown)


build_system.sh
```diff
@@ -67,10 +71,6 @@ exec_as_root() {
   docker exec $MOUNT_CONTAINER_ID "$@"
 }
 
-# Create host user in container (fixes namespace.so error)
-USERNAME=$(whoami)
```
andiradulescu, line 71: you need to replace missing $USERNAME in `docker exec -u $USERNAME $MOUNT_CONTAINER_ID "$@"`

andiradulescu:
Using the same solution also in https://github.com/commaai/agnos-builder/pull/247 even though the 20.04 image doesn't have user 1000, when we upgrade to mainline we might use 24.04 and it will break then without this change.

Dockerfile.sparsify
```diff
@@ -1,9 +1,18 @@
...
+ARG UID
+ARG GID
+
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && \
     apt-get install -y android-sdk-libsparse-utils && \
     rm -rf /var/lib/apt/lists/*
 
+RUN userdel -r `getent passwd ${UID} | cut -d : -f 1` > /dev/null 2>&1; \
```
andiradulescu, line 13: should not be done if UID is 0 (root) - just as a fail safe - e.g. https://github.com/commaai/agnos-builder/pull/247/commits/553b9b81ad7eea30321cf772d8a775cf3717e293

build_system.sh
robin-reckmann: Yes! Good catch. It actually works as is (at least when using zsh) because USERNAME is set by default. But better to explicitly set it. I'll take over your code.

Dockerfile.sparsify
robin-reckmann: Not sure if I get your motivation here correctly. But yes its not necessary since you would delete and recreate the same root user. Are there any other issues that could happen?

andiradulescu: I tried running it with sudo e.g. `sudo build_system.sh` and it failed with `useradd: user 'root' already exists` - probably userdel didn't work for root (the message was silenced and didn't check)

Issue #240, opened 2024-07-14
Switch from pyenv to uv


adeebshihadeh:
Similar to #232, but for managing the Python installation

eshaanrathi2:
@adeebshihadeh uv is a replacement for pip. i don't think it is equivalent for pyenv.
though, would it help if I migrate pip package installs to uv, perhaps making it faster?

adeebshihadeh:
```bash
batman:~$ uv python list
warning: `uv python list` is experimental and may change without warning.
cpython-3.12.4-linux-x86_64-gnu     <download available>
cpython-3.11.9-linux-x86_64-gnu     <download available>
cpython-3.11.4-linux-x86_64-gnu     .pyenv/versions/3.11.4/bin/python3
cpython-3.11.4-linux-x86_64-gnu     .pyenv/versions/3.11.4/bin/python
cpython-3.10.14-linux-x86_64-gnu    <download available>
cpython-3.9.19-linux-x86_64-gnu     <download available>
cpython-3.8.19-linux-x86_64-gnu     <download available>
cpython-3.8.10-linux-x86_64-gnu     /usr/bin/python3
cpython-3.8.10-linux-x86_64-gnu     /bin/python3
cpython-3.7.9-linux-x86_64-gnu      <download available>
batman:~$ uv python install 3.12
warning: `uv python install` is experimental and may change without warning.
Searching for Python versions matching: Python 3.12
Found existing installation for Python 3.12: cpython-3.12.3-linux-x86_64-gnu
batman:~$
```

coder351:
@adeebshihadeh I have prepared PR(https://github.com/commaai/agnos-builder/pull/248) which switches pyenv to uv for openpilot_python_dependencies.sh.

pantew869:
I've spent a couple of hours on this, dbus-python fails to build because meson can't find pkg-config files for python, and even if you fix this, the pkg-config files provided by the uv have incorrect paths in them.

PR #248, opened 2024-07-17
switch pyenv to uv

coder351:
resolves https://github.com/commaai/agnos-builder/issues/240

openpilot_python_dependencies.sh has been setup to use uv. Dockerfile.agnos still uses pyenv which can also be switched to use uv.

Issue #225, opened 2024-05-25
Reduce AGNOS system image size to <1GB


adeebshihadeh:
Convenience things, like `valgrind`, should be the last  to go. First up is anything we're just wasting space on, like package caches or AGNOS build-time dependencies.

If we have to start cutting these things, I think we'll want to setup an "AGNOS dev" package that will install all those extras together.

```
batman:output$ du -hs system.img
4.3G    system.img
```

`scripts/check_space.sh` is a nice tool for debugging this and will allow you to poke around like this:
![image](https://github.com/commaai/agnos-builder/assets/8762862/d3b9cafc-3dea-496f-ab17-c404e0dccabe)

jyoung8607:
![Screenshot_20240525-191830](https://github.com/commaai/agnos-builder/assets/46612682/03aa6803-1456-4587-a37e-4f7d86424010)

Not requesting a lock, that's an idea free for anyone's use.

andiradulescu:
Activated `strictatime` on my C3 and works nice, but it's more like a fine tuning approach, which I'll do after I clean stuff like this:

<img width="527" alt="image" src="https://github.com/user-attachments/assets/26b9ebd5-afde-41d3-a716-e90262a9e639">

`debs` folder in site-packages... :exploding_head: (most likely because of an `uv` bug)

After fixing the above, I will clean `pyproject.toml` to be similar to openpilot's `pyproject.toml`.

adeebshihadeh:
Locked to @andiradulescu

Doncuppjr:
I don't know that I ever seen quite such a sprawling mess, and I've been doin embedded for a long time. SELinux PAM, source code all in a image ?????

andiradulescu:
@Doncuppjr Thanks for the feedback! And I'm not being sarcastic. Myself I had close to zero experience with embedded, when I contributed.

If you're interested, you can take this bounty or even outline a plan to clean the "mess". The plan can then be broken into bounty-sized tasks.

Also comma.ai is hiring a Systems Engineer. Your embedded experience could be valuable - feel free to check out the opportunities at [comma.ai/jobs](https://comma.ai/jobs). The easiest way to get a job is doing one of the bounties.

Doncuppjr:
[...]

Issue #30, opened 2022-01-02
build python with optimizations


adeebshihadeh:
Seeing about a 10% improvement in controlsd with a quick test. Also worth trying to build capnproto with LTO and PGO.

For bounty, make sure to post before/after benchmarks.

pantew869:
> Also worth trying to build capnproto with LTO and PGO.
Should capnproto be profiled with openpilot unit-tests?

andiradulescu:
added https://github.com/commaai/agnos-builder/issues/353 for capnproto

PR #253, opened 2024-07-22
switch pyenv to uv 

coder351:
The PR is follow up of https://github.com/commaai/agnos-builder/pull/248 and resolves https://github.com/commaai/agnos-builder/issues/240

The DockerFile also needed to be switched to uv. The pyqt5 library uses pyqt5-qt5 which only has wheels for macos. pyqt6 has wheels for linux (arm64 and x86-64). so switched to pyqt6 and pyqt6-sip.
The dbus-python library also had missing dependencies libdbus-glib-1-dev , libgirepository1.0-dev. The pkg-config also was failing due to missing pc files for python 311. Package python3-dev and python3.12-dev installs pc files python-3.12.pc and python-3.11-embed.pc. The python-3.12.pc files can also be used for building python 3.11. Full build is passing without errors.


userspace/openpilot_dependencies.sh
```diff
@@ -3,6 +3,7 @@
 echo "Installing openpilot dependencies"
 
 # Install necessary libs
+add-apt-repository ppa:deadsnakes/ppa -y
```
adeebshihadeh, line 6: why is this needed if we’re using uv?

Dockerfile.agnos
```diff
@@ -78,24 +78,28 @@ RUN /tmp/agnos/openpilot_python_dependencies.sh
...
-    eval "$(pyenv virtualenv-init -)" && \
+RUN export PATH="$HOME/.cargo/bin:$PATH" && \
+    export UV_BIN="$HOME/.cargo/bin" && \
     export MAKEFLAGS="-j$(nproc)" && \
-    pip install pyqt5-sip==12.12.1 && \
-    pip install pyqt5==5.15.9 --verbose --config-settings --confirm-license= && \
-    pyenv rehash
+    source .venv/bin/activate && \
+    .venv/bin/python --version && \
+    uv pip install pyqt6-sip==13.8.0 && \
```
robin-reckmann, line 86:
```suggestion
    uv pip install pyqt5-sip==12.12.1 && \
```

Dockerfile.agnos
```diff
@@ -78,24 +78,28 @@ RUN /tmp/agnos/openpilot_python_dependencies.sh
...
+    uv pip install pyqt6==6.7.1 --verbose --config-settings="--confirm-license="
```
robin-reckmann, line 87:
```suggestion
    uv pip install pyqt5==5.15.9 --verbose --config-settings="--confirm-license=" --no-deps
```

robin-reckmann: pyqt5 can be installed from sources. I guess we should stay with pyqt5 for now.

robin-reckmann:
With your changes python and all packages from the pyproject.toml are installed to either /tmp/agnos or the home folder of the root user. These places are not accessible when you boot the image. Also userspace/files/profiles should be adapted accordingly.

adeebshihadeh:
Also needs to be PyQt5 since openpilot is on Qt5.

Dockerfile.agnos
coder351: I have reverted to pyqt5.

userspace/openpilot_dependencies.sh
coder351: the python3.12-dev package provided by the repo is not needed since uv python also provides .pc files for pkgconfig and the entry can be removed.

Dockerfile.agnos
coder351: I have reverted to pyqt5.

coder351:
> With your changes python and all packages from the pyproject.toml are installed to either /tmp/agnos or the home folder of the root user. These places are not accessible when you boot the image. Also userspace/files/profiles should be adapted accordingly.

@robin-reckmann I checked using ./scripts/check-space.sh and anything installed using uv pip including packages from pyproject.toml are stored in /.venv/lib/. I have also adapted the various services and userspace/files/profile to use uv python installation (/.venv/bin/python). 

I also noticed  gabrielfalcao/pyenv-action@2f49ca7587f9d0663d13f1147b78d3361417eaf7 is used by build_kernel.sh and is installing python2. Removing the github action causes the github build to fail. uv python provides provider version > 3.7, perhaps the necessary thing to do would be to also switch from python2 to python3. Also if possible, the generated system.img should be booted to test and confirm if the services are able to load correctly using uv based python.

andiradulescu:
@coder351 - `python2` in CI for `build_kernel.sh` has no connection with what happens inside `Dockerfile.agnos`. Just ignore it, the current kernel needs python2 to build.

Also, with https://github.com/commaai/agnos-builder/pull/247 `python2` is removed from CI.

I might be able to try out what you did here, after I get my PR merged.

coder351:
@adeebshihadeh The PR also resolves the https://github.com/commaai/agnos-builder/issues/30 (build python with optimizations). After installing UV, I can check the build flags and uv python is installed using --enable-optimizations

CONFIG_ARGS = "'--build=x86_64-unknown-linux-gnu' '--host=x86_64-unknown-linux-gnu' '--prefix=/install' '--with-openssl=/tools/deps' '--with-system-expat' '--with-system-libmpdec' '--without-ensurepip' '--with-readline=editline' '--enable-shared' '--enable-optimizations' '--with-lto`

For claiming the bounty I am also posting before and after benchmark for python, before took 5:14 and after took 2:51.

<details>

<summary>Before using --enable-optimizations</summary>
===================================================================== slowest 10 durations =====================================================================
17.09s call     selfdrive/car/tests/test_car_interfaces.py::TestCarInterfaces::test_car_interfaces_025_FORD_EXPLORER_MK6
16.01s call     selfdrive/car/tests/test_car_interfaces.py::TestCarInterfaces::test_car_interfaces_026_FORD_FOCUS_MK4
15.92s call     selfdrive/car/tests/test_car_interfaces.py::TestCarInterfaces::test_car_interfaces_027_FORD_F_150_LIGHTNING_MK1
15.88s call     selfdrive/car/tests/test_car_interfaces.py::TestCarInterfaces::test_car_interfaces_028_FORD_F_150_MK14
15.44s call     selfdrive/car/tests/test_car_interfaces.py::TestCarInterfaces::test_car_interfaces_024_FORD_ESCAPE_MK4
15.43s call     selfdrive/car/tests/test_car_interfaces.py::TestCarInterfaces::test_car_interfaces_029_FORD_MAVERICK_MK1
15.38s call     selfdrive/car/tests/test_car_interfaces.py::TestCarInterfaces::test_car_interfaces_023_FORD_BRONCO_SPORT_MK1
15.14s call     selfdrive/car/tests/test_car_interfaces.py::TestCarInterfaces::test_car_interfaces_030_FORD_MUSTANG_MACH_E_MK1
14.91s call     selfdrive/car/tests/test_car_interfaces.py::TestCarInterfaces::test_car_interfaces_034_GENESIS_G80
14.59s call     selfdrive/car/tests/test_car_interfaces.py::TestCarInterfaces::test_car_interfaces_099_KIA_FORTE
=============================================================== 208 passed in 314.95s (0:05:14) ================================================================
</details>

<details>

<summary>After using --enable-optimizations</summary>
using uv python '--enable-optimizations' '--with-lto'
===================================================================== slowest 10 durations =====================================================================
8.26s call     selfdrive/car/tests/test_car_interfaces.py::TestCarInterfaces::test_car_interfaces_192_VOLKSWAGEN_ATLAS_MK1
8.20s call     selfdrive/car/tests/test_car_interfaces.py::TestCarInterfaces::test_car_interfaces_150_SKODA_OCTAVIA_MK3
8.15s call     selfdrive/car/tests/test_car_interfaces.py::TestCarInterfaces::test_car_interfaces_112_KIA_OPTIMA_H_G4_FL
8.14s call     selfdrive/car/tests/test_car_interfaces.py::TestCarInterfaces::test_car_interfaces_198_VOLKSWAGEN_PASSAT_NMS
8.13s call     selfdrive/car/tests/test_car_interfaces.py::TestCarInterfaces::test_car_interfaces_191_VOLKSWAGEN_ARTEON_MK1
8.11s call     selfdrive/car/tests/test_car_interfaces.py::TestCarInterfaces::test_car_interfaces_195_VOLKSWAGEN_GOLF_MK7
8.06s call     selfdrive/car/tests/test_car_interfaces.py::TestCarInterfaces::test_car_interfaces_149_SKODA_KODIAQ_MK1
7.97s call     selfdrive/car/tests/test_car_interfaces.py::TestCarInterfaces::test_car_interfaces_199_VOLKSWAGEN_POLO_MK6
7.89s call     selfdrive/car/tests/test_car_interfaces.py::TestCarInterfaces::test_car_interfaces_197_VOLKSWAGEN_PASSAT_MK8
7.88s call     selfdrive/car/tests/test_car_interfaces.py::TestCarInterfaces::test_car_interfaces_151_SKODA_SUPERB_MK3
=============================================================== 208 passed in 171.89s (0:02:51) ================================================================
</details>

coder351:
@andiradulescu Thanks for helping test in the mean time, I suppose if anybody has instructions if possible on how to test booting the image (perhaps with QEMU) without the actual hardware would be helpful.

adeebshihadeh:
> @andiradulescu Thanks for helping test in the mean time, I suppose if anybody has instructions if possible on how to test booting the image without the actual hardware would be helpful.

I'd recommend preparing an openpilot branch and reaching out on Discord for someone to give it a quick test.

coder351:
@andiradulescu since PR (https://github.com/commaai/agnos-builder/pull/247) is merged, please do help test booting the image on physical device. Thanks.

andiradulescu:
The PR looks good, will boot it and let you know how it goes.

andiradulescu:
@coder351 can you push another commit that has [upload] inside the commit message - the commit can be empty - so it uploads the artifact for me to download and flash?

robin-reckmann:
Are we ok with moving the venv to /.venv ? I would prefer to keep it in /usr/local/venv

Dockerfile.agnos
robin-reckmann: Did the installation with uv like suggested above not work?

Dockerfile.agnos
```diff
@@ -78,25 +78,30 @@ RUN /tmp/agnos/openpilot_python_dependencies.sh
...
-    export PYENV_ROOT="/usr/local/pyenv" && \
-    eval "$(pyenv init -)" && \
-    eval "$(pyenv virtualenv-init -)" && \
+RUN export PATH="$HOME/.cargo/bin:$PATH" && \
+    export UV_BIN="$HOME/.cargo/bin" && \
+    source .venv/bin/activate && \
     cd /tmp/agnos/uv && \
     export PYOPENCL_CL_PRETEND_VERSION="2.0" && \
-    MAKEFLAGS="-j$(nproc)" UV_NO_CACHE=1 uv pip install --no-cache-dir --system .
+
```
robin-reckmann, line 96: Remove empty line to prevent the following docker warning:
WARN: NoEmptyContinuation: Empty continuation line

adeebshihadeh:
> Are we ok with moving the venv to /.venv ? I would prefer to keep it in /usr/local/venv

Let's just keep it the same.

robin-reckmann:
Hi @coder351 , I created a PR onto this branch with some modifications. Please have a look and merge / cherry-pick if you like. 

https://github.com/coder351/agnos-builder/pull/1

Summary:
- use uv instead of pip to install PyQt5
- move venv to /usr/local/venv
- remove pip shim
- remove pip installation
- remove explicit python installation (install it during venv creation instead)

robin-reckmann:
@adeebshihadeh Is pip still necessary somewhere else or can we remove it if it can be completely replaced with uv during the agnos image build?

coder351:
@robin-reckmann uv pip install fails to install pyqt5 from sources, the error I am getting is  No solution found when resolving dependencies. pip is already using uv based python and /usr/local/venv to install pyqt5 from sources so makes no difference to add uv in the front.

robin-reckmann:
Adding --no-deps should do the trick.

coder351:
@robin-reckmann using --no-deps does build on aarch64, but for x86_64 uv pip installs without the use of --no-deps.
Also pip is already using uv based python and install pyqt5 without the issue of dependencies. PyQt5 does not have wheels built for aarch64 but PyQt6 does have wheels built and uv installs PyQt6 without dependency issues. Perhaps leave the pip to use uv based python to install PyQt5 and when switching to PyQt6, also use uv pip instead of pip. I can add comment to explain why pip is used. Eitherways PyQt5 does build using pip or uv pip --no-deps, I would defer to @andiradulescu and @adeebshihadeh as to whether to go with uv pip --no-deps or pip (uv python)

robin-reckmann:
Ok, either is fine for me. We could also consider later to create the wheel and store it in the repo for faster installation.

coder351:
@andiradulescu I have pushed the empty commit for testing.

andiradulescu:
the [upload] tag can be inside your last commit message, no need for an empty commit everytime.

will test it soon.

andiradulescu:
Flashed the image, everything seems to work normal. Also ran `test_onroad`.

coder351:
@andiradulescu Thanks. @adeebshihadeh The PR has been tested and can be merged. Please note because pyQt5 wheel needs to be built each time the build time is increased. I am able to include pre-built wheel for pyQt5 in the PR to help reduce build time.

userspace/usr/comma/shims/pip
adeebshihadeh, line 1: how do I `pip install` now?

userspace/openpilot_dependencies.sh
```diff
@@ -56,6 +56,8 @@ apt-get install --no-install-recommends -yq \
     ocl-icd-opencl-dev \
     opencl-headers \
     pkg-config \
+    libdbus-glib-1-dev \
+    libgirepository1.0-dev \
```
adeebshihadeh, line 60: these are just build time dependencies of PyQt5? if so, these shouldn't end up in the final image. see how `agnos-compiler` is used for some other stuff

coder351: @adeebshihadeh not for PyQt5 but for dbus-python in pyproject.toml.

userspace/usr/comma/shims/pip
coder351: the shims were used by pyenv, the pip file was removed and also the shims folder. 
pip install can still be used using uv based python, e.g.
source $XDG_DATA_HOME/venv/bin/activate
pip install pyqt5==5.15.9 --verbose --config-settings="--confirm-license="

coder351: pip can still be used along with uv based python or system python. The shims folder and the pip file were used by pyenv which are no longer needed. To use pip install, activate the venv and call pip install e.g.
source /usr/local/venv/bin/activate
pip install pyQt5

robin-reckmann: I assume @adeebshihadeh refers to `pip install` on the device? That's actually a good point and we broke that functionality here (read-only filesystem). Lets bring this file back and re-enable` uv pip install` on the device.

coder351:
@robin-reckmann I wanted to request to put PR https://github.com/commaai/agnos-builder/pull/278 on hold while the pyenvtouv PR is merged because PRs like https://github.com/commaai/agnos-builder/pull/277 are duplicating the effort. The pyenv PR already handled the FromAsCasing warning. Thanks.

userspace/openpilot_dependencies.sh
coder351: @adeebshihadeh not for PyQt5 but for dbus-python in pyproject.toml. In total about 71 build time dependencies are getting installed besides the dbus-python libraries. The build time dependencies can be removed by calling purge in DockerFile but I recommend to handle build-time optimizations in another PR.

coder351:
@adeebshihadeh The script openpilot_dependencies.sh, is duplicated and run on agnos-compiler and agnos-base. 
https://github.com/commaai/agnos-builder/blob/360675b9357a9ec2dbfa80d9dd5220a9ad77a4b1/Dockerfile.agnos#L35
https://github.com/commaai/agnos-builder/blob/360675b9357a9ec2dbfa80d9dd5220a9ad77a4b1/Dockerfile.agnos#L73

The duplication is basically installing 71 libraries in agnos-base. The openpilot_dependencies.sh should only run on agnos-compiler and only necessary packages should be installed in agnos-base. After some testing I found out of 71 packages only 3 packages are necessary to build pyproject.toml and the rest of the DockerFile. I have placed the 3 packages namely portaudio19-dev, opencl-headers, gcc-arm-none-eabi in openpilot_python_dependencies.sh.
By removing the duplication, the build size is also reduced. In regards to removing build time dependencies i.e packages ending with -dev, I ran some tests and found the command apt-get purge lib*-dev is able to remove -dev packages and conserve about 0.5 gb of disk space from the final image. The command fails when run in github runner due to frontend-lock. I have also put back shims/pip point pip to uv based venv. The PR can be merged and the removal of dev packages would reduce the agnos system image size which can be discussed in https://github.com/commaai/agnos-builder/issues/225.

coder351:
@adeebshihadeh The script openpilot_dependencies.sh, is duplicated and run on agnos-compiler and agnos-base. 
https://github.com/commaai/agnos-builder/blob/360675b9357a9ec2dbfa80d9dd5220a9ad77a4b1/Dockerfile.agnos#L35
https://github.com/commaai/agnos-builder/blob/360675b9357a9ec2dbfa80d9dd5220a9ad77a4b1/Dockerfile.agnos#L73

The duplication is basically installing 71 libraries in agnos-base. The openpilot_dependencies.sh should only run on agnos-compiler and only necessary packages should be installed in agnos-base. After some testing I found out of 71 packages only 3 packages are necessary to build pyproject.toml and the rest of the DockerFile. I have placed the 3 packages namely portaudio19-dev, opencl-headers, gcc-arm-none-eabi in openpilot_python_dependencies.sh.
By removing the duplication, the build size is also reduced. In regards to removing build time dependencies i.e packages ending with -dev, I ran some tests and found the command apt-get purge lib*-dev is able to remove -dev packages and conserve about 0.5 gb of disk space from the final image. The command fails when run in github runner due to frontend-lock. I have also put back shims/pip, pointing PIP_PATH to uv based pip in venv. The PR can be merged and the removal of dev packages would reduce the agnos system image size which can be discussed in https://github.com/commaai/agnos-builder/issues/225.

userspace/usr/comma/shims/pip
robin-reckmann, line 1: Permission are not the same anymore. Should be 755

Dockerfile.agnos
```diff
@@ -211,9 +213,8 @@ RUN echo -n "10.1" > /VERSION
 # ################## #
 # #### Cleaunup #### #
 # ################## #
-
 RUN rm -rf /usr/share/icons/* && \
     rm -rf /var/lib/apt/lists/* && \
     rm -rf /home/$USERNAME/.cache && \
     rm -rf /root/.cache && \
-    apt-get clean
+    apt-get clean
```
robin-reckmann, line 220: Also the newline should probably be preserved. (I guess its a setting of your IDE/text editor). Same for the other files where this happened.

Dockerfile.agnos
```diff
@@ -26,7 +26,7 @@ RUN /tmp/agnos/base_setup.sh
 # ################## #
 # #### Compiler #### #
 # ################## #
-FROM agnos-base as agnos-compiler
+FROM agnos-base AS agnos-compiler
```
robin-reckmann, line 29: These should not be part of this PR as they are unrelated to the uv introduction.

Dockerfile.agnos
```diff
@@ -69,34 +69,36 @@ COPY --from=agnos-compiler-capnp /tmp/capnproto.deb /tmp/capnproto.deb
...
-    eval "$(pyenv init -)" && \
-    eval "$(pyenv virtualenv-init -)" && \
-    export MAKEFLAGS="-j$(nproc)" && \
-    pip install pyqt5-sip==12.12.1 && \
-    pip install pyqt5==5.15.9 --verbose --config-settings --confirm-license= && \
-    pyenv rehash
+ENV XDG_DATA_HOME="/usr/local"
+ENV PATH="$XDG_DATA_HOME/.cargo/bin:$PATH"
+
+COPY ./userspace/files/PyQt5-5.15.9-cp38-abi3-manylinux_2_17_aarch64.whl /tmp/agnos/
```
robin-reckmann, line 82: From this [comment](https://github.com/commaai/agnos-builder/pull/278#issuecomment-2264599903) the way to go is to still build PyQt5 for now but make sure its somewhere in the build where other changes to the dockerfile dont break the cache. Also its unrelated to the uv introduction, so could we do this in another PR and remove it here?

userspace/files/profile
```diff
@@ -32,32 +32,10 @@ export QT_QPA_PLATFORM="wayland-egl"
...
-    command pyenv "$command" "$@";;
-  esac
-}
-
-# custom shims
-export PATH="/usr/comma/shims:${PATH}"
 
+export PATH="/usr/local/.cargo/bin:$PATH"
+export XDG_DATA_HOME="/usr/local"
+source $XDG_DATA_HOME/venv/bin/activate
```
robin-reckmann, line 38: @adeebshihadeh This will activate the venv for all sessions on the device. It is a different behavior compared to now, where you have to use `pyenv exec`. Is this ok?

Dockerfile.agnos
coder351: @robin-reckmann please see previous comment https://github.com/commaai/agnos-builder/pull/253#discussion_r1698344560. The warning removal was requested and the FromAsCasing is also part of warnings.

robin-reckmann: Ah yes, that comment was only for the newly introduced warning. But this part is already fixed on master anyway.

Dockerfile.agnos
coder351: according to docker build cache invalidation general rules,
"During cache lookup, cache is invalidated if the file metadata has changed for any of the files involved."
the docker cache should not bust as long as the wheel file metadata is not modified. I have already built the wheel and setup in dockerfile so lets keep using the pre-built PyQt5 wheel also the build time optimization is much needed.

Dockerfile.agnos
coder351: done.

userspace/usr/comma/shims/pip
coder351: done.

userspace/files/profile
coder351: @robin-reckmann pyenv exec would still use python "3.11.4", when run using pyenv exec
https://github.com/commaai/agnos-builder/blob/360675b9357a9ec2dbfa80d9dd5220a9ad77a4b1/userspace/files/profile#L36
unless the idea is to use different python versions across sessions, then activating venv ensures the python version 3.11.4 is consistent across sessions.

Dockerfile.agnos
coder351: according to docker build cache invalidation general rules,
"During cache lookup, cache is invalidated if the file metadata has changed for any of the files involved."
the docker cache should not bust as long as the wheel file metadata is not modified. I have already built the wheel and setup in dockerfile and using the pre-built PyQt5 wheel also provides the much needed build time optimization. Unless updating to pyqt6, I suggest to keep using the pre-built wheel.

adeebshihadeh:
With 24.04, we will no longer need to install a custom python (#262).

Gonna close this since 24.04 will be merged soon, but you're good to claim the bounty since the work was done.

coder351:
@adeebshihadeh Thanks, regarding the bounty, please highlight exactly how much I should be claiming, also please note regarding issue https://github.com/commaai/agnos-builder/issues/30, I had posted before and after benchmarks, the uv based python is already using optimized python. So if the issue is still valid, perhaps the python portion of the bounty can also be factored in.

PR #260, opened 2024-07-25
remove timezonefinder

andiradulescu:
Previously removed in openpilot https://github.com/commaai/openpilot/issues/32662

`timezonefinder` is in the top 5 python deps (for size):

image


adeebshihadeh:
We should be able to just make the openpilot package a dependency and relock to pull in the openpilot dependencies.

andiradulescu:
That's a really nice idea, will try it!

andiradulescu:
Closing in favor of https://github.com/commaai/agnos-builder/pull/265

PR #264, opened 2024-07-26
update flash scripts to always flash active slot

andiradulescu:
- moving to `system.img` from `system-skip-chunks.img` (in order to remove `system-skip-chunks.img` later)
- `flash_kernel.sh` with `flash_system.sh` always flashes the active slot


adeebshihadeh:
Neither the existing stuff or this change is quite correct. Let's just flash to the active slot.

andiradulescu:
Updated to always flash the active slot.

PR #265, opened 2024-07-26
Split out openpilot and agnos python dependencies

andiradulescu:
Resolves https://github.com/commaai/agnos-builder/issues/23


andiradulescu:
this branch is driving my car right now 😍

andiradulescu:
```
du -sh system.img
3.4G	system.img
```

adeebshihadeh:
> For #225
> 
> ```
> du -sh system.img
> 3.4G	system.img
> ```

Would be cool to make a script that does some basic analysis and run it in CI.

andiradulescu:
I'll add a step in the CI at the end.

Dockerfile.agnos
```diff
@@ -88,13 +88,16 @@ RUN export PATH="/usr/local/pyenv/bin:/usr/local/pyenv/shims:$PATH" && \
     pyenv rehash
 
 # Install openpilot python packages
-RUN mkdir -p /tmp/agnos/uv
-COPY ./userspace/pyproject.toml ./userspace/uv.lock /tmp/agnos/uv
+RUN mkdir -p /tmp/agnos/uv-agnos /tmp/agnos/uv-openpilot
+COPY ./userspace/pyproject.toml ./userspace/uv.lock /tmp/agnos/uv-agnos
+RUN cd /tmp/agnos/uv-openpilot && \
+    curl -sO https://raw.githubusercontent.com/commaai/openpilot/master/pyproject.toml && \
+    curl -sO https://raw.githubusercontent.com/commaai/openpilot/master/uv.lock
```
adeebshihadeh, line 95: if i build this in 2 years, it's gonna pull the uv.lock from the latest openpilot.

ideally, we can check in a `uv.lock` and refresh that as needed.

adeebshihadeh (changes requested):

Dockerfile.agnos
andiradulescu: done

andiradulescu:
Compared to master:
```
  4.3G	/
  1.1G	/usr/local/venv
  854M	/usr/lib/aarch64-linux-gnu
  338M	/usr/lib/llvm-18
  260M	/usr/lib/gcc
  155M	/usr/lib/arm-linux-gnueabihf
  331M	/usr/bin
  328M	/usr/share
  166M	/usr/comma
```

sync_openpilot_dependencies.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+set -e
+
+# Make sure we're in the correct spot
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null && pwd)"
+cd $DIR
+
+cd userspace/uv
+curl -sSo openpilot/pyproject.toml https://raw.githubusercontent.com/commaai/openpilot/master/pyproject.toml
+curl -sSo uv.lock https://raw.githubusercontent.com/commaai/openpilot/master/uv.lock
```
adeebshihadeh, line 10: we shouldn't need it's uv.lock

userspace/uv/pyproject.toml
```diff
@@ -0,0 +1,20 @@
+[project]
+name = "agnos"
+version = "0.1.0"
+authors = [{name = "Vehicle Researcher", email = "user@comma.ai"}]
+requires-python = "~=3.11"
+description = ""
+dependencies = [
+  "openpilot",
+  "openpilot[testing]",
```
adeebshihadeh, line 9: ideally we want all the openpilot groups

userspace/uv/pyproject.toml
```diff
@@ -0,0 +1,20 @@
+[project]
+name = "agnos"
+version = "0.1.0"
+authors = [{name = "Vehicle Researcher", email = "user@comma.ai"}]
+requires-python = "~=3.11"
+description = ""
+dependencies = [
+  "openpilot",
```
adeebshihadeh, line 8: why can't we use the git url?

andiradulescu: uv has a bug:

- if I put "openpilot@git+https://github.com/commaai/openpilot" in dependencies it will crash trying to pull submodules. I saved a [log](https://gist.github.com/andiradulescu/6b94eae07bd297e74668d42a53b85d08).
- works fine with pip install, so it's something related to uv.

That's why I decided to curl master pyproject.toml. Also, this is much faster, since cloning the whole openpilot repo takes precious time.

userspace/uv/pyproject.toml
andiradulescu: "testing" and "dev" make sense. I don't see the point for "docs" though. what do you think?

andiradulescu: although:
- the "dev" dependencies are not useful for normal users
- the "testing" must be there for testing, but again not useful for normal users 

In the cleanup phase or when we want a minimal image, we can simply keep just "openpilot" and install dev+testing+docs with one command (unminimize).

adeebshihadeh: yeah, that's fine as long as we have that unminimize command and it's in the MOTD

andiradulescu:
@adeebshihadeh `uv` somehow _decided_ to not use the system env anymore to install, thus no python packages installed in this last image - fixing this in https://github.com/commaai/agnos-builder/pull/341 + `uv sync` wasn't a good idea

PR #268, opened 2024-07-28
update bash scripts to #!/usr/bin/env bash

andiradulescu:
Mostly for macOS.

A good explaination [here](https://stackoverflow.com/a/58494728/639708).

Same thing for:
- openpilot https://github.com/commaai/openpilot/pull/33111
- panda https://github.com/commaai/panda/pull/1987

\* left `userspace` alone, for now, since it only runs on device

### DIFF
--- a/build_kernel.sh
+++ b/build_kernel.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e
 
 DEFCONFIG=tici_defconfig
--- a/build_kernel_headers.sh
+++ b/build_kernel_headers.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e
 
 DEFCONFIG=tici_defconfig
--- a/build_system.sh
+++ b/build_system.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e
 
 UBUNTU_BASE_URL="http://cdimage.ubuntu.com/ubuntu-base/releases/20.04/release"
--- a/flash_all.sh
+++ b/flash_all.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e
 
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null && pwd)"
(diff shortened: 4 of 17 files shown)


adeebshihadeh:
Can we setup pre-commit linting for this?

andiradulescu:
We can setup pre-commit for this, but I'm starting to think to simply use:
```
#!/bin/bash
```
than
```
#!/usr/bin/env bash
```
and never:
```
#!/usr/bin/bash
```
More reading [here](https://unix.stackexchange.com/questions/29608/why-is-it-better-to-use-usr-bin-env-name-instead-of-path-to-name-as-my).

If we want pre-commit for scripts, we could also use:
- [shellcheck-py/shellcheck-py](https://github.com/shellcheck-py/shellcheck-py): runs [shellcheck](https://www.shellcheck.net/) on your scripts
- [openstack/bashate](https://github.com/openstack/bashate): code style enforcement for bash programs

(they don't check if you use `#!/bin/bash` or `#!/usr/bin/env bash` or `#!/usr/bin/bash`)

I kindly ask for more opinions: @adeebshihadeh  @maxime-desroches

---
FYI, this PR started from the fact that macOS doesn't have `bash` in `/usr/bin/` but only in `/bin`.

If a newer version of `bash` is installed on macOS, then `#!/usr/bin/env bash` will use that one instead of the one in `/bin`.

adeebshihadeh:
Adding a bash linter sounds good, though make sure to keep the enforcing similar to our other linters. We generally only want to error on things that are clearly bad patterns or bugs.

andiradulescu:
Will see what I can do about the bash linter.

Any preference towards `#!/usr/bin/env bash` vs `#!/bin/bash` ? Linted shebang?

adeebshihadeh:
No preference, just want to enforce it's all the same with a linter.

adeebshihadeh:
https://github.com/commaai/openpilot/issues/32761 is relevant here. We probably want to do some shebang linting.

maxime-desroches:
I have not found an already existing linter that does this (especially for python3 vs python in the shebang). 
@andiradulescu Did you managed to find a linter that does that?

andiradulescu:
No, I didn't. Maybe make our own linter(s)? That's what I want to do regarding the shebang.

PR #271, opened 2024-07-28
no internal

adeebshihadeh:


.gitmodules
```diff
@@ -1,12 +1,3 @@
-[submodule "edk2_tici"]
```
adeebshihadeh, line 1: FYI @robbederks

We can setup a private builder repo with these, but I don't think it'll be too bad like this.

Issue #272, opened 2024-07-28
Build + push full release in CI


adeebshihadeh:
The goal is to have CI output a full manifest you can drop in an openpilot branch for easy testing. Here's an example AGNOS update: https://github.com/commaai/openpilot/pull/32373.

The workflow will then be:
* push your changes to a PR
* bump the AGNOS version
* CI gives you a manifest to drop in an openpilot branch
* share on Discord for easy testing, or try it yourself

andiradulescu:
As explained [here](https://github.com/commaai/agnos-builder/pull/275#issuecomment-2266681174), GitHub artifacts are always zip files. As for Releases/Prereleases (and the required tags) they will spam the agnos-builder repo [a lot](https://github.com/actions/upload-artifact/issues/27#issuecomment-581381318).

So, the only ways I can see this working:
- use GH releases and write a zip stream decompressor (in openpilot) specifically for the zip format used in GitHub CI
- use Azure (with secrets added in the repo) to upload usual xz files packaged by `package_ota.py`

Azure seems the cleanest way.

PR #275, opened 2024-07-30
Ship non-sparse system image in OTA

andiradulescu:
Prep for:
- https://github.com/commaai/agnos-builder/issues/5
- https://github.com/commaai/agnos-builder/issues/272

Depends on [commaai/flash](https://github.com/commaai/flash) to not flash `-skip-chunks` image [anymore](https://github.com/andiradulescu/commaai-flash/commit/d8c7b52d09d02845205bdca5774814d7e9180ae6) (**tested**).


andiradulescu:
As next step here, I'm thinking on comparing the flash times for:

- raw
- sparsified
- sparsified with skip chunks

on:
- fastboot
- edl

same:
- host
- cable
- device

and come up with a table.

adeebshihadeh:
Looked over this again, and I think we did already ship the necessary openpilot support for this. Just need to set `sparse` to False in the manifest and make sure the openpilot updater works.

https://github.com/commaai/openpilot/blob/cafca4f89176855585835ff7c67dea5a9b08ca23/system/hardware/tici/agnos.py#L162

andiradulescu:
Yes, I looked before on `agnos.py` but didn’t actually test it (at all) yet.

Also, I wasn’t sure you wanted to remove both sparsified images. Now I understand that `yes`. So, I will update this PR.

One thing to keep in mind. GH artifacts always archives artifacts as zip files and it can’t be changed. So archiving as xz files will still get them zipped. And zip files can’t use a stream decompressor since they are not meant for this, the header is at the [end of the file](https://stackoverflow.com/a/29376408/639708).

So, CI builds with artifacts can’t be used for a direct update/flash.

But, GH Releases support files up to 2GB. We can use a tag in the commit message that triggers the creation of a Prerelease with agnos.json and all boot.img.xz + system.img.xz.

andiradulescu:
Had to reduce the raw image size since openpilot was complaining with the 10G image:
```
INFO:root:Downloading and writing system
INFO:root:Installing system: 1 - Total written: 108003328 - Chunk size: 1048576 - Memory usage: 17.8%
INFO:root:Installing system: 2 - Total written: 214958080 - Chunk size: 1048576 - Memory usage: 18.1%
Traceback (most recent call last):
  File "/data/openpilot/system/hardware/tici/agnos.py", line 331, in <module>
    flash_agnos_update(args.manifest, target_slot_number, logging, standalone=True)
  File "/data/openpilot/system/hardware/tici/agnos.py", line 287, in flash_agnos_update
    flash_partition(target_slot_number, partition, cloudlog, standalone)
  File "/data/openpilot/system/hardware/tici/agnos.py", line 250, in flash_partition
    extract_compressed_image(target_slot_number, partition, cloudlog)
  File "/data/openpilot/system/hardware/tici/agnos.py", line 165, in extract_compressed_image
    for chunk in f(downloader):
  File "/data/openpilot/system/hardware/tici/agnos.py", line 88, in noop
    yield f.read(1024 * 1024)
          ^^^^^^^^^^^^^^^^^^^
  File "/data/openpilot/system/hardware/tici/agnos.py", line 41, in read
    out = self.decompressor.decompress(compressed)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
MemoryError: Unable to allocate output buffer.
finished with  1
```
The space on `/dev/root` (`/`) can be expanded with one simple command:
```
sudo resize2fs $(findmnt -n -o SOURCE /)
```

andiradulescu:
In one sentence, the raw image is reduced to 5G and flashing works perfectly fine.

@adeebshihadeh is it worth it to fix openpilot to accept the 10G image? or we merge it as it is?

adeebshihadeh:
I suggest we proceed like this:
- [ ] fix openpilot so we can expand the size later on
- [ ] make sure to do the resize in the apt/pip hook when we remount (it shouldn't resize unless you're already modifying the system image)

I put up a bounty to finish this side of it and locked to you #5.

andiradulescu:
- [x] openpilot [flash fix](https://github.com/commaai/openpilot/pull/33320)
- [x] resize2fs in apt/fix hook

tools/simg2dontcare.py
```diff
@@ -1,76 +0,0 @@
-#!/usr/bin/env python3
-import argparse
```
adeebshihadeh, line 2: we need to keep skip-chunks around for flash.comma.ai. this PR should just be for the openpilot OTA image

build_system.sh
```diff
@@ -14,10 +14,8 @@ BUILD_DIR="$DIR/build"
 OUTPUT_DIR="$DIR/output"
 
 ROOTFS_DIR="$BUILD_DIR/agnos-rootfs"
-ROOTFS_IMAGE="$BUILD_DIR/system.img.raw"
+ROOTFS_IMAGE="$BUILD_DIR/system.img"
 ROOTFS_IMAGE_SIZE=10G
```
adeebshihadeh, line 18: once this is changed to 5G, needs a comment explaining why (link here) and explain that we have to until most devices have the updater fix to bump it back up

tools/simg2dontcare.py
andiradulescu: got it, wasn’t sure we wanted to keep it or not.

- I assumed that if we need to push an AGNOS 10.2 update before the 11 release, we could just fork from 10.1 and push any patches on a 10.2 hotfix branch. used this “workflow” before in other projects when an intermediary update was needed while working at the next major release (that wasn’t ready for release).

- 2nd point about comma flash, we can just fake the “alt” image to be the same image as the main (raw) one - so comma flash doesn’t crash looking for the “alt” image.

these two points could make sense so that we don’t spam GitLab LFS with two 800 MB files per push.

but I agree with you that having both images is a much cleaner (even if getting close to the 10 GB limit on GitLab LFS)

build_system.sh
andiradulescu: since resize2fs enlarge is already in place, I’m planning to leave 10G here, and just use “resize2fs -M” (Shrink the filesystem to the minimum size.) at the end.

since flashing a raw 10G image vs the minimum or 5G one takes longer. I can benchmark the time needed.

the minimum image or the 5G one flashed without memory errors.

adeebshihadeh: I added the 5G back, otherwise we could silently ship an image that's too big. It's unlikely but we should be very careful with anything involving the updater.

PR #278, opened 2024-08-01
[no merge] System build time improvements

robin-reckmann:
PR for experiments on CI system build time


robin-reckmann:
Baseline: 30m18s

robin-reckmann:
Replace apt-get with apt-fast: 27m49s

robin-reckmann:
Combine apt package installation in base_setup: No improvement ( ~28min).

robin-reckmann:
Use pre-compiled PyQt5 wheel: 15m15s ( - 12m30s)

adeebshihadeh:
> Use pre-compiled PyQt5 wheel: 15m15s ( - 12m30s)

Can we refactor the build such that the docker cache will never bust unless we change some PyQt stuff?

robin-reckmann:
Yes, that should be possible. With this and refactoring the other compile scripts to use a minimized build environment and run them in parallel to the base_setup etc. we should get below 10min (if we dont count the PyQt5 compilation that should always be cached then)

adeebshihadeh:
Yeah, I'd count that for the bounty as long as it's just PyQt5.

robin-reckmann:
On the other hand, I see it as completely valid to use this specific whl instead of compiling from sources. Reasoning: There is no source code modification and if we ever switch to newer versions (i.e. PyQt6) pre-compiled wheels are anyway available and this whole pip installation step can be moved into the pyproject.toml.

andiradulescu:
Took a look over your last build and the biggest hoggers are:
| Script | Duration (seconds) |
|--------|-------------------:|
| base_setup.sh | 206.1 |
| compile-qtwayland5.sh | 201.6 |
| compile-capnp.sh | 161.4 |
| compile-modemmanager.sh | 159.9 |
| compile-ffmpeg.sh | 192.6 |
| openpilot_python_dependencies.sh | 96.8 |
| uv pip install | 83.3 |
| openpilot_dependencies.sh | 42.6 |
| install_extras.sh | 14.4 |

Ignored what's taking ~10s or less.

No more pyqt compiling! Only taking 11.0s to install the whl.

robin-reckmann:
Use uv instead of pyenv (@coder351 s PR): 12m15s ( - 1m15s)

coder351:
> On the other hand, I see it as completely valid to use this specific whl instead of compiling from sources. Reasoning: There is no source code modification and if we ever switch to newer versions (i.e. PyQt6) pre-compiled wheels are anyway available and this whole pip installation step can be moved into the pyproject.toml.

@robin-reckmann If needed I am able to include the wheel. Please feel free to make suggestions in the PR (https://github.com/commaai/agnos-builder/pull/253)

> Yeah, I'd count that for the bounty as long as it's just PyQt5.

@adeebshihadeh In regards to the bounty, I am able to include the PyQt5 wheel if needed :) I was not clear if uploading wheels to the repo is possible/feasible thing to do. Perhaps upgrading to PyQt6 makes sense and I can also raise PR to bump openpilot to use PyQt6. Upgrading to PyQt6 would make the pyqt5 uv step redundant in Docker and also improve build times. I am curious to find any particular reason not to upgrade to PyQt6 given the wheels are prebuilt and also reduce the Docker build step. Thanks.

robin-reckmann:
To be continued in https://github.com/commaai/agnos-builder/pull/296

robin-reckmann:
To be continued in https://github.com/commaai/agnos-builder/pull/296

PR #282, opened 2024-08-05
build and push full release in CI to Azure

andiradulescu:
For https://github.com/commaai/agnos-builder/issues/272

As a one-time setup, it needs an `AZURE_CREDENTIALS` secret in GitHub, that has the permission to push to a separate container in Azure - named it `agnosupdate-ci`.

To create [AZURE_CREDENTIALS](https://learn.microsoft.com/en-us/cli/azure/azure-cli-sp-tutorial-1?tabs=bash#create-a-service-principal-with-role-and-scope):
```
az ad sp create-for-rbac --name "GitHub CI" \
                         --role "Storage Blob Data Contributor" \
                         --scopes /subscriptions/<subscription-id>/resourceGroups/<resource-group>/providers/Microsoft.Storage/storageAccounts/commadist/blobServices/default/containers/agnosupdate-ci
```

The return of this command needs to be saved as the `AZURE_CREDENTIALS` secret.

\* `<subscription-id>` and `<resource-group>` need to be replaced with the respective values
\** the scope defined above restricts it only to the `agnosupdate-ci` container


andiradulescu:
@adeebshihadeh - I did a proposal for https://github.com/commaai/agnos-builder/issues/272. You can also take a look over the last [run](https://github.com/commaai/agnos-builder/actions/runs/10254682889). What do you think?

Currently it uploads the files to my Azure Personal account - for testing.

.github/workflows/build.yaml
```diff
@@ -76,3 +76,26 @@ jobs:
       with:
         name: system.img
         path: output/system.img
+
+    - name: Package and OTA push
+      if: "contains(env.LAST_COMMIT_MESSAGE, '[release]')"
+      env:
+        AGNOS_UPDATE_URL: https://openpilot.blob.core.windows.net/agnosupdate-ci/${{ github.run_id }}
+        # TODO: switch to comma credentials: https://learn.microsoft.com/en-us/cli/azure/ad/sp?view=azure-cli-latest#az-ad-sp-create-for-rbac-examples
+        AZURE_CREDENTIALS: '{"appId": "fd6e3414-7fee-4bf3-bd85-dd36d3d03bd2", "password": "Xs-8Q~uaKQaqhGdggDyLmTkwjIa3lNzh2B9AKboR", "tenant": "499a1903-3119-470b-9dd4-d4f436586f25"}'
```
adeebshihadeh, line 85: We're working remove Azure from openpilot's CI since it's extra infra that only we have access to (you can't fork or make a PR to an azure bucket).

Let's use https://github.com/commaai/ci-artifacts like openpilot does

adeebshihadeh (changes requested):

.github/workflows/build.yaml
adeebshihadeh: https://github.com/commaai/openpilot/blob/d00a53983089be6bed0e037474abfea823f2fc3e/.github/workflows/ui_preview.yaml#L41

adeebshihadeh: yes, `CI_ARTIFACTS_DEPLOY_KEY`. I'm ok trading off some speed for externalized infrastructure.

adeebshihadeh:
Looks good, remaining TODOs:
* Azure -> commaai/ci-artifacts
* this will replace the GH actions artifact upload, so we can remove that
* it should be gated on `[upload]`
* would be cool if it commented on PRs with a one-liner to get a matching openpilot branch: "Push this to an openpilot branch like this: ..."

Once the system build is fast and the output image is small, we'll remove the `[upload]` trigger and always package/upload.

andiradulescu:
Closing in favor of https://github.com/commaai/agnos-builder/pull/287

PR #286, opened 2024-08-07
Remove apt cache for each layer

robin-reckmann:
Remove cached apt-related files after each build step that includes apt package installations. This reduces the size of the final docker image from 6.61GB to 5.85GB.
Note: It does NOT affect the size of the final system.img

### DIFF
--- a/userspace/base_setup.sh
+++ b/userspace/base_setup.sh
@@ -106,8 +106,6 @@ apt-fast install --no-install-recommends -yq \
     wireless-tools \
     zlib1g-dev
 
-rm -rf /var/lib/apt/lists/*
-
 # Allow chrony to make a big adjustment to system time on boot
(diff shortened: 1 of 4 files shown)


andiradulescu:
Wouldn't it be better to not remove `/var/lib/apt` at all between steps and just delete it in the last `Dockerfile.agnos` step?

Another optimization I see would be to call `apt update` just once at the beginning e.g. in `base_setup.sh`.

andiradulescu:
> see here https://docs.docker.com/build/building/best-practices/ 

Very nice write up!

adeebshihadeh:
It's not clear to me this tradeoff is worth it. Where is the individual layer size a problem?

robin-reckmann:
This is also done in https://github.com/commaai/agnos-builder/pull/308 in a better way. Closing.

PR #288, opened 2024-08-07
build and push full release in CI on [upload]

andiradulescu:


andiradulescu:
@adeebshihadeh same permission error: https://github.com/commaai/agnos-builder/actions/runs/10293101463/job/28488769124

andiradulescu:
ok, so comparing an openpilot run https://github.com/commaai/openpilot/actions/runs/10289970225/job/28478858166 shows that it has a "ssh-key":
image
vs here where there is no "ssh-key":
image
I'm thinking of two possibilities:
- there is some environment set in openpilot or something special [here](https://github.com/commaai/openpilot/settings/secrets/actions) or [here](https://github.com/commaai/openpilot/settings/environments) - can you check (yet again) @adeebshihadeh ?
- this secret doesn't work on namespace.so runners - I will ask on Namespace Discord

andiradulescu:
Namespace responded, they don’t influence secrets at all, which makes sense. buy they suggested we check the repos whitelist here: https://github.com/organizations/commaai/settings/secrets/actions/CI_ARTIFACTS_DEPLOY_KEY

andiradulescu:
Found the problem, we need two workflows..
- one workflow that runs on "pull_request" - building (this does not have access to secrets)
- another workflow on completion of the previous - uploads to ci-artifacts and comments in PR (this one has access to secrets)

This means that we still need to upload and download artifacts between these two workflows..

Explained here (ReceivePR.yml and CommentPR.yml): https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

adeebshihadeh:
Added `CI_ARTIFACTS_GITLAB_DEPLOY_KEY` for https://gitlab.com/commaai/ci-artifacts. We can push big LFS stuff there.

andiradulescu:
- pushes to GitHub
- 14min xz both images
- 14sec artifact upload with 0 compression

doing the xz in the first workflow which uses namespace for heavy lifting. the second workflow uses GH ubuntu-latest.

next thing switch to GitLab.

should we merge this and see if the workflow works and make a separate PR for GitLab?

or finish it here?

adeebshihadeh:
Let's merge, then we can optimize if needed later.

adeebshihadeh:
Failed for #304: https://github.com/commaai/agnos-builder/actions/runs/10343772270/job/28628349425

andiradulescu:
I ran it in another testing repo, meanwhile, and it had a lot of issues. It was pretty impossible to get this right blindly.

When I'm done testing, I'll tag in you in the new PR.

PR #291, opened 2024-08-08
agnos rootfs statistics in CI

andiradulescu:


robin-reckmann:
Would be nice if this step could automatically write a comment with the image size etc. into the PR. Also the folders in the list are bit wonky. Maybe you could list all folders above a certain threshold but then ignore their parent folders.

adeebshihadeh:
Agree with @robin-reckmann. There's just too much going on down there too.

What's the most info we can convey in 5-10 lines? It's generally better to provide an incomplete but succinct summary than try to fit everything in.

andiradulescu:
Yes, I thought the same, to be in the comment from https://github.com/commaai/agnos-builder/pull/288

As for 5-10 lines, I’ll do what @robin-reckmann suggested.

andiradulescu:
Is this good enough?

It says total size, python env size, (apt) libs size, and other bigger /usr/* subdirs.

### AGNOS Stats:
```
4.0G    /
1.5G    /usr/local/pyenv
735M    /usr/lib/aarch64-linux-gnu
294M    /usr/lib/gcc
156M    /usr/lib/arm-linux-gnueabihf
311M    /usr/share
178M    /usr/bin
166M    /usr/comma
```

adeebshihadeh:
LGTM!

PR #295, opened 2024-08-09
[no merge yet] System python 3.12 and uv venv

robin-reckmann:
[NO MERGE] (still has to be rebased onto master once 24.04 is merged)
(based on @coder351 s PR: https://github.com/commaai/agnos-builder/pull/253)
Use system python 3.12 instead of installing 3.11
Replace pyenv venv with uv venv


robin-reckmann:
Going with uv installed python instead

PR #299, opened 2024-08-09
[NO MERGE] Test Dockerfile.agnos checking

robin-reckmann:
Test to check if issues in the Dockerfile.agnos are properly detected and an error is thrown.


adeebshihadeh:
Good to close?

robin-reckmann:
Yes. Will close

PR #302, opened 2024-08-10
Build PyQt5 wheel in parallel

robin-reckmann:
Build PyQt5 wheel in parallel to other compilation steps and install from wheel later.
This reduces the overall Docker build time and makes better use of the Docker build cache.

Tests: Builds and runs on C3. Openpilot compiles, GUI works. (Not tested on-road)

### DIFF
--- a/Dockerfile.agnos
+++ b/Dockerfile.agnos
@@ -40,6 +40,12 @@ FROM agnos-base AS agnos-compiler
 RUN apt-fast update && apt-fast install --no-install-recommends -yq checkinstall
 
 # Individual compiling images
+FROM agnos-compiler AS agnos-compiler-pyqt5
+COPY ./userspace/openpilot_python_dependencies.sh /tmp/agnos/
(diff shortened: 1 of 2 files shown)


Dockerfile.agnos
```diff
@@ -29,6 +29,12 @@ COPY ./userspace/openpilot_dependencies.sh /tmp/agnos/
 RUN /tmp/agnos/openpilot_dependencies.sh
 
 # Individual compiling images
+FROM agnos-compiler AS agnos-compiler-pyqt5
+COPY ./userspace/openpilot_python_dependencies.sh /tmp/agnos/
+RUN /tmp/agnos/openpilot_python_dependencies.sh
```
adeebshihadeh, line 34: why is this done twice?

adeebshihadeh: can we copy the rest of the dependencies out of here too?

adeebshihadeh:
@robin-reckmann want to finish this one up? Looks ready to go aside from the double python deps.

PR #262, opened 2024-07-25
Ubuntu 24.04

andiradulescu:
Part of https://github.com/commaai/openpilot/issues/32386
- 24.04 old canonical: https://github.com/commaai/agnos-builder/pull/221
- 24.04 original PR: https://github.com/commaai/agnos-builder/pull/188

---
## Release Checklist
- [x] [`test_onroad`](https://github.com/commaai/openpilot/blob/master/selfdrive/test/test_onroad.py) passes
- [x] Wi-Fi: lists networks and connects
- [x] Modem: connects to cell network
- [ ] Image sizes haven't increased **(increased to 4.5G from 4.0G from merged master)**
- [x] Sounds work
- [x] `python` is our python, not system version
- [x] Clean openpilot build: `scons -c && scons -j8`
- [x] Factory reset
  - [x] from openpilot menu
  - [x] tapping on boot
  - [ ] corrupt userdata
- [ ] Color calibration
  - [ ] from /persist/comma/
  - [ ] directly from panel over sysfs
- [x] Clean setup: factory reset -> install openpilot -> openpilot works
- [ ] AGNOS update works on warm boot
  - [ ] previous -> new
  - [ ] new -> previous

### ABL
- [x] Fastboot USB enumeration
- [x] Boot time hasn't regressed (3.8s -> 3.7s)

### XBL
- [ ] Display init works in cold and hot temperatures
- [x] Boot time hasn't regressed (2.4s)

#### Networking
- [x] Continue button disabled when no connection
- [x] Forget/connect to wifi

#### Custom URL
- (a) Not a real URL (e.g. `comma`, `abc123`, `...`)
  - [x] "Ensure the entered URL is valid"
  - [x] Start over
  - [x] Reboot device
- (b) Website but not an installer URL (e.g. `github.com`, `comma.ai`, `installer.comma.ai`)
  - [x] "No custom software found at this URL."
- (c) Valid installer URL (e.g. `openpilot.comma.ai`)
  - [x] Download successful (comma logo or installer appears)
  - [x] `/tmp/installer_url` should contain the installer URL

PR #303, opened 2024-08-10
pyenv to uv based python and venv

coder351:
resolves https://github.com/commaai/agnos-builder/issues/240 and https://github.com/commaai/agnos-builder/issues/30

UV based python uses optimized build
CONFIG_ARGS = "'--build=x86_64-unknown-linux-gnu' '--host=x86_64-unknown-linux-gnu' '--prefix=/install' '--with-openssl=/tools/deps' '--with-system-expat' '--with-system-libmpdec' '--without-ensurepip' '--with-readline=editline' '--enable-shared' '--enable-optimizations' '--with-lto`

UV based python is set to 3.12.3.
PyQt5 uses prebuilt wheels.

TODOs
- [x] I see uv has the optimizations enabled by default, but can we explicitly enable them so we're not relying on defaults?
- [x] PyQt5: either need a script to build/pull the wheel or build/install in the Docker build again
- [x] final test on device
- [x] openpilot runs
- [x] python, python3, pip, pip3, scons all work in a shell logged in as comma


adeebshihadeh:
Marking as draft until TODOs are resolved

coder351:
@adeebshihadeh 

TODOs
- [x] I see uv has the optimizations enabled by default, but can we explicitly enable them so we're not relying on defaults?

In the uv-python code base, when uv python install is called, the download-metadata.json is used to determine the download link of the indygreg based python standalone builds, e.g.

https://github.com/astral-sh/uv/blob/15dfb660ab7f1ccca7e57c985968a18a85bb1d86/crates/uv-python/download-metadata.json#L13

"url": "https://github.com/indygreg/python-build-standalone/releases/download/20240814/cpython-3.12.5%2B20240814-aarch64-unknown-linux-gnu-install_only_stripped.tar.gz",

Taking the build alias from the url 'aarch64-unknown-linux-gnu' and opening the corresponding release list and then looking up aarch64-unknown-linux-gnu,

https://github.com/indygreg/python-build-standalone/blob/main/src/release.rs#L97

In aarch64-unknown-linux-gnu, install_only_suffix is set to 'lto' which uses link time optimizations.
install_only_suffix: "lto",

So any time when building for aarch64, the install_only_suffix is set to lto and safe to assume the aarch64 build would have lto, otherwise uv does not provide any explicit parameter when installing python.

- [x] PyQt5: either need a script to build/pull the wheel or build/install in the Docker build again
Going with build/install in docker build again, I have added build flag rebuildpyqt5=0, if the flag is set to 0, prebuilt wheel is used and if set to 1, the pyqt5 wheel is rebuilt.

coder351:
@andiradulescu from what I can gather the remaining todos need physical device, please do quick test when available.
- [ ] final test on device
- [ ] openpilot runs
- [ ] python, python3, pip, pip3, scons all work in a shell logged in as comma

Dockerfile.agnos
```diff
@@ -83,28 +80,34 @@ RUN cd /tmp && apt-get -o Dpkg::Options::="--force-overwrite" install -yq ./capn
 COPY --from=agnos-compiler-ffmpeg /tmp/ffmpeg.deb /tmp/ffmpeg.deb
 RUN cd /tmp && apt-get -o Dpkg::Options::="--force-overwrite" install -yq ./ffmpeg.deb
 
-# Use pre-compiled PyQt5 wheel
-COPY --from=agnos-compiler-pyqt5 /tmp/PyQt5-5.15.9/PyQt5-5.15.9-cp38-abi3-manylinux_2_17_aarch64.whl /tmp/PyQt5-5.15.9-cp38-abi3-manylinux_2_17_aarch64.whl
+ENV XDG_DATA_HOME="/usr/local"
+ENV PATH="$XDG_DATA_HOME/.cargo/bin:$PATH"
+ARG rebuildpyqt5=0
```
adeebshihadeh, line 85: any reason not to do this every time now that it's parallel?

adeebshihadeh (changes requested):

userspace/openpilot_python_dependencies.sh
```diff
@@ -1,21 +1,14 @@
...
 
-echo "Installing python for openpilot"
+echo "installing uv..."
 
-# Install pyenv
-export PYENV_ROOT="/usr/local/pyenv"
-curl https://pyenv.run | bash
-export PATH="$PYENV_ROOT/bin:$PATH"
-eval "$(pyenv init -)"
+export XDG_DATA_HOME="/usr/local"
```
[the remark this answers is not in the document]
coder351: @adeebshihadeh The XDG_DATA_HOME only sets the uv venv location (usr/local) otherwise the default location would fall to current directory (.venv).

Dockerfile.agnos
coder351: Considering the todo item PyQt5: either need a script to build/pull the wheel or build/install in the Docker build again,
Even with parallel builds, seems building wheel every time would slow down the build. I have put build flag rebuildpyqt5=0 when set to 0 uses predefined wheel when set to 1 builds the wheel. I suppose other options could include to stay with building the wheel every time but then using prebuilt wheel becomes redundant.

userspace/openpilot_dependencies.sh
```diff
@@ -56,6 +56,8 @@ apt-fast install --no-install-recommends -yq \
     ocl-icd-opencl-dev \
     opencl-headers \
     pkg-config \
+    libdbus-glib-1-dev \
+    libgirepository1.0-dev \
```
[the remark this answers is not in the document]
coder351: done.

robin-reckmann:
Boots and compiles:

```
comma@comma-xxxxxxxx:/data/openpilot$ which python
/usr/local/venv/bin/python

comma@comma-xxxxxxxx:/data/openpilot$ which python3
/usr/local/venv/bin/python3

comma@comma-xxxxxxxx:/data/openpilot$ which pip
/usr/local/venv/bin/pip

comma@comma-xxxxxxxx:/data/openpilot$ which pip3
/usr/local/venv/bin/pip3

comma@comma-xxxxxxxx:/data/openpilot$ which scons
/usr/local/venv/bin/scons

comma@comma-xxxxxxxx:/data/openpilot$ python
Python 3.12.3 (main, Apr 15 2024, 17:09:49) [GCC 6.3.0 20170516] on linux
Type "help", "copyright", "credits" or "license" for more information.
```

userspace/openpilot_dependencies.sh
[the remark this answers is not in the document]
robin-reckmann: Are these necessary? Everything works without them and they were not installed before.

Dockerfile.agnos
```diff
@@ -83,29 +80,34 @@ RUN cd /tmp && apt-get -o Dpkg::Options::="--force-overwrite" install -yq ./capn
 COPY --from=agnos-compiler-ffmpeg /tmp/ffmpeg.deb /tmp/ffmpeg.deb
 RUN cd /tmp && apt-get -o Dpkg::Options::="--force-overwrite" install -yq ./ffmpeg.deb
 
-# Use pre-compiled PyQt5 wheel
-COPY --from=agnos-compiler-pyqt5 /tmp/PyQt5-5.15.9/PyQt5-5.15.9-cp38-abi3-manylinux_2_17_aarch64.whl /tmp/PyQt5-5.15.9-cp38-abi3-manylinux_2_17_aarch64.whl
+ENV XDG_DATA_HOME="/usr/local"
+ENV PATH="$XDG_DATA_HOME/.cargo/bin:$PATH"
```
robin-reckmann, line 84:
```suggestion
ARG XDG_DATA_HOME="/usr/local" 
ARG PATH="$XDG_DATA_HOME/.cargo/bin:$PATH"
```

robin-reckmann, line 84:
```suggestion
ARG PATH="$XDG_DATA_HOME/.cargo/bin:$PATH"
```

userspace/openpilot_python_dependencies.sh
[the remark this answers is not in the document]
robin-reckmann: Here it doesn't affect anything except providing the installation part for the venv. But we should avoid usage it /etc/profile as this will make it Into the envvars and might have some influence.

userspace/files/profile
```diff
@@ -32,32 +32,10 @@ export QT_QPA_PLATFORM="wayland-egl"
...
-  *)
-    command pyenv "$command" "$@";;
-  esac
-}
-
-# custom shims
-export PATH="/usr/comma/shims:${PATH}"
 
+export PATH="/usr/local/.cargo/bin:$PATH"
+export XDG_DATA_HOME="/usr/local"
```
robin-reckmann, line 37:
```suggestion
```

userspace/files/profile
```diff
@@ -32,32 +32,10 @@ export QT_QPA_PLATFORM="wayland-egl"
...
+source $XDG_DATA_HOME/venv/bin/activate
```
robin-reckmann, line 38:
```suggestion
source /usr/local/venv/bin/activate
```

userspace/openpilot_dependencies.sh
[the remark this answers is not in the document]
coder351: PR https://github.com/commaai/agnos-builder/pull/262 added deps libgirepository1.0-dev and libdbus-1-dev to build libqmi and modemManager and the deps in openpilot_python_dependencies.sh are no longer needed.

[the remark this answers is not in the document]
coder351: @adeebshihadeh PR https://github.com/commaai/agnos-builder/pull/262 added deps libgirepository1.0-dev and libdbus-1-dev to build libqmi and modemManager and the deps in openpilot_python_dependencies.sh are no longer needed.

coder351:
@adeebshihadeh The remaining Todos and comments are resolved.

Dockerfile.agnos
robin-reckmann: How did you build the wheel? rebuildpyqt5 does not actually rebuild it as in saving an updated whl file, but just chooses between your file and building from scratch.

coder351: @robin-reckmann rebuildpyqt5 would either select the prebuilt wheel or build again from source and then save the file for use. Perhaps the better alternative as suggested by @adeebshihadeh is to rebuild wheel every time but doing so makes the prebuilt wheel redundant so perhaps drop the prebuilt wheel and rebuild wheel every time.

coder351: @robin-reckmann rebuildpyqt5 would either select the prebuilt wheel or build again from source and then save the file for use. Perhaps the better alternative as suggested by @adeebshihadeh is to rebuild wheel every time but doing so makes the prebuilt wheel redundant so perhaps remove the prebuilt wheel and rebuild wheel every time.

robin-reckmann: Yes, ideally we would just use the whl file and have a completely separated script to build it on demand. My only issue with your solution is, that the whl file is not actually saved outside of the docker build context. And the ARG is not expose to the build script.

robin-reckmann: @adeebshihadeh If the .whl file were available on the PyPI server, we probably wouldn't have an issue using it, or?(Similar to not re-building all the deb files). Given that we don't even modify it (for now) and it saves a considerable amount of build time, I would again propose to completely separate out the whl compile and use that to update if necessary.

adeebshihadeh: Let's build it every time. I already explained my rationale in the other PR

adeebshihadeh (changes requested):
PyQt5 needs to be cleaned up still, other than that LGTM

coder351:
@adeebshihadeh I have the refactored PyQt5 to rebuild each time. The build time was 5mins which includes the docker cache.

userspace/compile-pyqt5.sh
```diff
@@ -1,16 +1,13 @@
 #!/bin/bash -e
 
-# Use pyenv venv
-export PATH="/usr/local/pyenv/bin:/usr/local/pyenv/shims:$PATH"
-export PYENV_ROOT="/usr/local/pyenv"
-eval "$(pyenv init -)"
-eval "$(pyenv virtualenv-init -)"
+export XDG_DATA_HOME="/usr/local"
+export CARGO_HOME="$XDG_DATA_HOME/.cargo"
+source $XDG_DATA_HOME/venv/bin/activate
```
robin-reckmann, line 5:
```suggestion
source /usr/local/venv/bin/activate 
```

userspace/compile-pyqt5.sh
```diff
@@ -1,16 +1,13 @@
 #!/bin/bash -e
 
-# Use pyenv venv
-export PATH="/usr/local/pyenv/bin:/usr/local/pyenv/shims:$PATH"
-export PYENV_ROOT="/usr/local/pyenv"
-eval "$(pyenv init -)"
-eval "$(pyenv virtualenv-init -)"
+export XDG_DATA_HOME="/usr/local"
```
robin-reckmann, line 3: Remove these lines. CARGO_HOME is not needed here.

robin-reckmann (changes requested):

robin-reckmann:
pip install does current not work.  `ERROR: Could not install packages due to an OSError: [Errno 30] Read-only file system: '/usr/local/venv/lib/python3.12/site-packages/<package_name>`. Somethings not working with the shim. @andiradulescu Could you have a look? Otherwise I can check tomorrow.

robin-reckmann:
Also uv does currently not have a shim. I think we should have one if we ever intend to use uv on device.

userspace/files/profile
```diff
@@ -32,32 +32,9 @@ export QT_QPA_PLATFORM="wayland-egl"
...
-
-  case "$command" in
-  activate|deactivate|rehash|shell)
-    eval "$(pyenv "sh-$command" "$@")";;
-  *)
-    command pyenv "$command" "$@";;
-  esac
-}
-
-# custom shims
```
robin-reckmann, line 58: We need the custom shim for pip. Please put back this two lines.

coder351:
@robin-reckmann I checked the build logs and compile-pyqt5.sh ran without any errors. I have added back the custom shims export path in userspace/files/profile.

userspace/compile-pyqt5.sh
```diff
@@ -1,10 +1,7 @@
...
```
robin-reckmann, line 3: This one is also not necessary.

coder351: done.

coder351:
@robin-reckmann any other issues or concerns otherwise the PR is ready.

coder351:
@robin-reckmann PR is ready if no other issues or concerns please confirm for @adeebshihadeh.

adeebshihadeh:
This should've been a straightforward PR, but it's taken a lot of people's time now. Bounty is unlocked, @robin-reckmann you want to pick it up?

coder351:
@adeebshihadeh the pr has been ready only needs testing on uv device, the op uv crash has nothing to do with the uv pr. Please allow testing time, I also put benchmark for python optimizations issue. Please reopen the PR and @robin and I can finalize the details on adding shim.

coder351:
@adeebshihadeh the pr has been ready only needs testing on uv device, the op ui crash has nothing to do with the uv pr. Please allow testing time, I also put benchmark for python optimizations issue. Please reopen the PR and @robin and I can finalize the details on adding shim.

PR #305, opened 2024-08-12
fixed build and push full release in CI 

andiradulescu:
This one works fine with GitHub. Keeping it as a draft until I get it working on GitLab. A couple of things to note:
- The approach is different from the one used in openpilot for screenshots - instead of spawning another workflow that sleeps and waits for the build workflow to finish, this one executes on completion of the build workflow
- For now, I'm just pushing `system.img.xz` and `boot.img.xz`, since `agnos.json` won't be used in comma flash (that currently uses `-skip-chunks`) and just don't want to waste LFS space (see below).
- I didn't specify anything related to increasing the AGNOS version, yet.

A test run can be seen here:
- https://github.com/andiradulescu/debug-github-actions-commit-message/pull/1 (scroll all the way to the bottom)
- https://github.com/andiradulescu/debug-github-actions-commit-message/actions
- https://github.com/andi-radulescu-contributor/ci-artifacts/tree/agnos-builder/pr-1

A couple of gotchas I found during testing:
- you can't push files over 100MB on git, files over 100 MB can be pushed only to LFS (GitHub and GitLab)
- when you rewrite the branch, you replace the LFS files, the old files remain in LFS and you can't delete them.. I found this for [GitHub](https://stackoverflow.com/questions/34579211/how-to-delete-a-file-tracked-by-git-lfs-and-release-the-storage-quota) - you basically need to recreate the repo to release the LFS space taken. I will see if this is the same for GitLab.

@adeebshihadeh if you want to merge it as it is, keep in mind what I said about GitHub LFS. I’m planning to have it ready for GitLab, by today.


adeebshihadeh:
Let's wait for GitLab then

andiradulescu:
GitLab is working!

Didn't want to move VERSION in this PR (wanted to do it in another), but I needed a simple way to pass VERSION in the PR comment. Comment looks like [this](https://github.com/andiradulescu/debug-github-actions-commit-message/pull/2).

What do you think?

Issue #10, opened 2021-11-08
power drop monitor: log timestamp to param


pd0wm:
When the power drop monitor triggers, append the current timestamp (seconds since boot) to a param. Openpilot can take care of making sure this gets cloudlogged and ends up in the qlogs next boot.

PR #308, opened 2024-08-13
Use separate cache for apt

robin-reckmann:
Move apt stuff to separate docker cache. This cache is reusable even if the Docker cache of the step itself breaks.

### DIFF
--- a/Dockerfile.agnos
+++ b/Dockerfile.agnos
@@ -14,21 +14,29 @@ RUN set -xe
 
 ARG USERNAME=comma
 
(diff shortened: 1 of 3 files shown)


robin-reckmann:
apt install step does not have to download packages, but uses cache instead:
```#10 73.07 debconf: delaying package configuration, since apt-utils is not installed
#10 73.20 0 upgraded, 457 newly installed, 1 to remove and 0 not upgraded.
#10 73.20 Need to get 0 B/166 MB of archives.
#10 73.20 After this operation, 727 MB of additional disk space will be used.```

robin-reckmann:
Todo: Check concurrent access of multiple runners

robin-reckmann:
Go with this for local execution and use namespace cache volumes for CI instead.

adeebshihadeh:
Won't be relevant once we switch to void!

PR #316, opened 2024-08-17
Extract apt cache for re-use in next build

robin-reckmann:
to be filled


robin-reckmann:
Potential downside is that docker cache is busted more often due to changes from even other PRs to the cache. With the high download speeds in the CI this doesnt seem to be that beneficial.

PR #317, opened 2024-08-18
image creation cleanup

andiradulescu:
Missed a `\` in `$DIR/tools/simg2dontcare.py \$TMP_SPARSE \$TMP_SKIP;` and also did some cleanup.

Testing to see it works fine locally, since CI has some hiccup.

### DIFF
--- a/Dockerfile.builder
+++ b/Dockerfile.builder
@@ -19,7 +19,6 @@ RUN apt-get update && \
     ccache \
     libcap2-bin \
     android-sdk-libsparse-utils \
-    e2fsprogs \
     && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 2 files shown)


andiradulescu:
Made more tests with `resize2fs -M` and I’m starting to not like it anymore. The size of the 10G image is reduced to ~7G.

Running it on the 5G, does nothing. Not really what I wanted.

I’m thinking on two options (both not using resize2fs for shrinking anymore):
- just use a fixed value e.g. 5G (actual data is around 4.6-4.7G)
- get the tar size (exported from docker) and add 100-300MB of free space

What do you think @adeebshihadeh?

adeebshihadeh:
I think we just stick with 5G, then bump it back up in ~6 months once the fix is out to most devices.

andiradulescu:
Agree. I will clean this PR then.

andiradulescu:
done

PR #331, opened 2024-08-30
Default to buildkit for kernel building

robbederks:
Without this, it takes the full directory in context, which is >30s on my machine.

### DIFF
--- a/build_kernel.sh
+++ b/build_kernel.sh
@@ -24,6 +24,7 @@ fi
 
 # Setup kernel build container
 echo "Building agnos-meta-builder docker image"
+export DOCKER_BUILDKIT=1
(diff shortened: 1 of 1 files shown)


andiradulescu:
BuildKit is the default since Docker 23. Are you using an older Docker version?

robbederks:
> BuildKit is the default since Docker 23. Are you using an older Docker version?

Yes apparently (20.10.8), seems like that's the version with ubuntu 20.04

andiradulescu:
@adeebshihadeh should we put `export DOCKER_BUILDKIT=1` back until everyone moves to 24.04?

If yes, we should also add it back in `build_system.sh`.

robbederks:
Also added it there. Think we merge it for now, and can revert once we move our workstations over to 24.04

PR #332, opened 2024-09-02
build_system: fail when ubuntu base is updated

andiradulescu:


adeebshihadeh:
This now silently updates the base image. The whole problem here is that the other failure is silent.

All we need to do here is make the docker build fail appropriately and update the URL. Base image updates should be explicit, even if necessary.

adeebshihadeh:
Propose that we download the `SHA256SUMS` file and ensure ours matches.

andiradulescu:
When you say "and update the URL" what do you mean? to print the new image url that we should commit manually?

adeebshihadeh:
Doesn't even have to be printed by the script (it's not worth the lines to do that). Just link to the page and we can update it manually whenever this happens.

andiradulescu:
Ok, done.

SHA256SUMS
```diff
@@ -0,0 +1,6 @@
+5b0ffb1e31d4b85f5a2214f105d1377ba616c453b38ff46c278dc6472b21e2d8 *ubuntu-base-24.04.1-base-amd64.tar.gz
```
adeebshihadeh, line 1: we shouldn't need to check this in

andiradulescu: done

PR #338, opened 2024-09-03
Add docker buildx version check

robin-reckmann:
Older docker versions don't have the `docker build --check` option.  Add version check, that skips the Dockerfile check when version is older than v0.15.0.


adeebshihadeh:
Discussed with @robbederks, he'll update his workstation!

PR #347, opened 2024-09-06
Cleanup llvm and qemu packages

andiradulescu:
Removes llvm/clang 18 and qemu-user and keeps llvm/clang 17.

### DIFF
--- a/Dockerfile.agnos
+++ b/Dockerfile.agnos
@@ -203,6 +203,10 @@ RUN cd /usr/lib/gcc/arm-none-eabi/* && \
     rm -rf arm/ && \
     rm -rf thumb/nofp thumb/v6* thumb/v8* thumb/v7+fp thumb/v7-r+fp.sp
 
+# TODO: make these packages not get installed in the first time
+# qemu-user also removes libglib2.0-dev:armhf libgudev-1.0-dev:armhf libinput-dev:armhf libwacom-dev:armhf
+RUN apt-fast update && apt-fast remove -y qemu-user:armhf
+
 # keep this last
 RUN ldconfig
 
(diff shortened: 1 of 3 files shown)


adeebshihadeh:
Why the older llvm?

userspace/base_setup.sh
```diff
@@ -108,6 +107,19 @@ apt-fast install --no-install-recommends -yq \
     wireless-tools \
     zlib1g-dev
 
+# keep LLVM 17 (not 18 as in 24.04, since mesa/rusticl/etc seems to use 17)
+# clang & llvm metapackages install LLVM 18 (which is not installed anymore)
+apt-fast install --no-install-recommends -yq \
+    clang-17 \
+    llvm-17
```
adeebshihadeh, line 114: I would worry about this later. we really don't want to add hacks now for future concerns. they'll be easy enough to change when we know we need them

andiradulescu: I don’t like this change, also, but 24.04 mesa depends right now on llvm-17. we can’t remove mesa since we need it for software rendering and also in mainline.

installing llvm 24.04 meta package installs llvm-18.

so, because of this we have both llvm-17 and llvm-18 installed, taking an extra ~150 MB of space.

we can avoid removing and keep both, until mesa updates to llvm-18, and llvm-17 will not get installed anymore (without any changes).

andiradulescu: @robin-reckmann what’s your take on this? I’m curious about your opinion.

Dockerfile.agnos
```diff
@@ -219,7 +223,7 @@ COPY ./userspace/files/apt.conf /etc/apt/apt.conf
 COPY VERSION /VERSION
 
 # ################## #
-# #### Cleaunup #### #
```
robin-reckmann, line 222: Most important cleaunup

andiradulescu:
I will close this for now and revisit later

PR #348, opened 2024-09-06
remove fonts

andiradulescu:
\- 94 MB


adeebshihadeh:
This breaks setup, reset, etc.

andiradulescu:
Ouch, weird that I didn’t noticed that upfront.

PR #352, opened 2024-09-06
Use debhelper tools for dependency compilation and packaging

robin-reckmann:
Use debhelper tools for creation of the .deb packages rather than old checkinstall.


andiradulescu:
I will run test_onroad on this, before merge.

robin-reckmann:
Restart in #408.

PR #354, opened 2024-09-07
Build independent packages first

robin-reckmann:
Move package compilation to the beginning of the main dockerfile. This parts can now run in parallel to the main installation up until the compiled packages are installed.


adeebshihadeh:
What does this improve? I generally find splitting up Dockerfiles to be pretty annoying since you have to manage each separately.

robin-reckmann:
> What does this improve? I generally find splitting up Dockerfiles to be pretty annoying since you have to manage each separately.

Ok, so then I would move the compilers to the top of the existing Dockerfile.

robin-reckmann:
Compile time test: https://github.com/commaai/agnos-builder/pull/359
min/max/avg : 8m46s / 9m02s / 8m54s

andiradulescu:
Also shared this on Discord - I would create "agnos-base" from the tar.gz file instead of "ubuntu:24.04". God knows what differences might be (now or in the future) between them.

Besides this, LGTM!

robin-reckmann:
I dont have a strong preference, which base to use. I compared the compiled binaries and they are identical. "agnos-base" is still using the tar.gz. Only the compiler is user docker ubuntu

PR #364, opened 2024-09-12
Compile only necessary PyQt5 modules

robin-reckmann:
Currently only QtCore and QtWidgers are in use. Disable all remaining modules to minimize compile time.


adeebshihadeh:
FYI, @lukasloetkolben we'll have to check in a wheel to use PyQt5 once the new AGNOS is out.

robin-reckmann:
Completely removed PyQt5. So we dont need this PR anymore.

Issue #318, opened 2024-08-19
NTP sync sometimes not instant


adeebshihadeh:
We should consider implementing a small time syncing service ourselves. It's super important we grab the time as soon as we can.

We use chrony right now: https://github.com/commaai/agnos-builder/blob/37a76e0950ff204722a1e2c45b289af09b70a4f0/userspace/base_setup.sh#L112

Issue #315, opened 2024-08-15
Boot time profiling script


adeebshihadeh:
For boot time, the first priority is anything needed to engage openpilot, and the second is auxiliary stuff a user would notice (wifi, modem, etc.)

Before we start improving boot time, we need a good test. Previous attempts have consistently hit long tail race conditions that led to regressions. This is a more detailed version of https://github.com/commaai/agnos-builder/issues/110#issuecomment-1445487013. 

Requirements:
* Records when a service (e.g. Weston) or peripheral (e.g. DSP) is ready
* Reports if something failed to come up
* Must capture everything openpilot uses and a user would notice: display, GPU, DSP, cameras, modem, Wi-Fi, etc.
* Has a mode to run in a loop and record stats

This script should answer the questions:
* how long until I can engage?
* how long until I can SSH into the device?

Doncuppjr:
`systemd-analyze plot` ?

PR #369, opened 2024-09-14
replace chrony with timesyncd

andiradulescu:
Resolves https://github.com/commaai/agnos-builder/issues/318
- removes chrony and replaces it with [systemd-timesyncd](https://wiki.archlinux.org/title/Systemd-timesyncd)
- makes time syncing faster than chrony and when network configuration or connectivity changes
- installed build and tested both with internet and without internet on boot - time is synced almost instantly after network connectivity is up

```
$ journalctl -u systemd-timesyncd
Aug 08 14:51:13 comma systemd[1]: Starting systemd-timesyncd.service - Network Time Synchronization...
Aug 08 14:51:13 comma (imesyncd)[2819]: systemd-timesyncd.service: ProtectHostname=yes is configured, but the kernel does not support UTS na>
Aug 08 14:51:13 comma systemd[1]: Started systemd-timesyncd.service - Network Time Synchronization.
Aug 08 14:51:13 comma systemd-timesyncd[2819]: No network connectivity, watching for changes.
Aug 08 14:54:47 comma-cb80a5fa systemd-timesyncd[2819]: Network configuration changed, trying to establish connection.
Sep 15 06:40:29 comma-cb80a5fa systemd-timesyncd[2819]: Contacted time server 185.125.190.57:123 (ntp.ubuntu.com).
Sep 15 06:40:29 comma-cb80a5fa systemd-timesyncd[2819]: Initial clock synchronization to Sun 2024-09-15 06:40:29.697339 UTC.
```

### DIFF
--- a/userspace/base_setup.sh
+++ b/userspace/base_setup.sh
@@ -58,7 +58,6 @@ apt-fast install --no-install-recommends -yq \
     build-essential \
     bzip2 \
     curl \
-    chrony \
     cpuset \
     dfu-util \
     evtest \
@@ -110,9 +109,6 @@ apt-fast install --no-install-recommends -yq \
     zlib1g-dev
 
 rm -rf /var/lib/apt/lists/*
-
-# Allow chrony to make a big adjustment to system time on boot
-echo "makestep 0.1 3" >> /etc/chrony/chrony.conf
 
 # Create dirs
 mkdir /data && chown $USERNAME:$USERNAME /data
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Once we get to https://github.com/commaai/agnos-builder/issues/315, I can see all sorts of checks being add if the script is written generically enough. Then, we can easily test the reliability of these things in a loop.

Issue #339, opened 2024-09-03
Clear `/data/tmp` on boot


adeebshihadeh:

andiradulescu:
Should `/data/tmp` exist anymore, then? Can't we just rename all `/data/tmp` references to `/tmp` and use `/tmp` instead ?

PR #371, opened 2024-09-15
clear /data/tmp on boot

andiradulescu:
Resolves https://github.com/commaai/agnos-builder/issues/339

Alternative (by mounting to `/tmp` variant):
```
# setup /data/tmp
rm -rf /data/tmp
ln -s /tmp /data/tmp
```

Tested both variants on device by editing `/usr/comma/comma.sh` manually, touching some files in `/data/tmp` then restart and investigate the results.

### DIFF
--- a/userspace/usr/comma/comma.sh
+++ b/userspace/usr/comma/comma.sh
@@ -61,8 +61,8 @@ elif ! mountpoint -q /data; then
 fi
 
 # setup /data/tmp
+rm -rf /data/tmp
 mkdir -p /data/tmp
-rm -rf /data/tmp/*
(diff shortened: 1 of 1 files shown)


userspace/usr/comma/comma.sh
```diff
@@ -60,6 +60,10 @@ elif ! mountpoint -q /data; then
   fi
 fi
 
+# setup /data/tmp
+mkdir -p /data/tmp
+rm -rf /data/tmp/*
```
adeebshihadeh, line 65: this won’t clear dot files

PR #377, opened 2024-09-18
B01 wlan config

robbederks:


robbederks:
still needs tuning, but we can do that later

Issue #374, opened 2024-09-16
fstab warning on v11


adeebshihadeh:
Noticed this in the main tmux pane
```
waiting for weston
weston ready after 5s
mount: (hint) your fstab has been modified, but systemd still uses
       the old version; use 'systemctl daemon-reload' to reload.
Removed "/etc/systemd/system/multi-user.target.wants/ssh-param-watcher.path".
Removed "/etc/systemd/system/multi-user.target.wants/ssh-param-watcher.service".
Disabling 'ssh-param-watcher.service', but its triggering units are still active:
ssh-param-watcher.path
mount: (hint) your fstab has been modified, but systemd still uses
       the old version; use 'systemctl daemon-reload' to reload.
```

PR #379, opened 2024-09-18
make timesync persistent

andiradulescu:
Fixes https://github.com/commaai/agnos-builder/issues/374 by:
- making the default system time to build time (by touching /usr/lib/clock-epoch) thus making the system time > /etc/fstab mtime (to prevent the fstab warning)
- symlinked timesyncd clock so that the last time set is set first at next boot
- validated by doing a couple of reboots and checking if 'systemctl daemon-reload' is requested
```
$ journalctl -u systemd-timesyncd
Sep 18 13:53:23 comma systemd[1]: Starting systemd-timesyncd.service - Network Time Synchronization...
Sep 18 13:53:23 comma (imesyncd)[3060]: systemd-timesyncd.service: ProtectHostname=yes is configured, but the kernel does not support UTS namespaces, ignoring namespace setup.
Sep 18 21:09:51 comma systemd-timesyncd[3060]: System clock time unset or jumped backwards, restored from recorded timestamp: Wed 2024-09-18 21:09:51 UTC
Sep 18 21:09:51 comma systemd[1]: Started systemd-timesyncd.service - Network Time Synchronization.
Sep 18 21:10:16 comma-cb80a5fa systemd-timesyncd[3060]: Network configuration changed, trying to establish connection.
Sep 18 21:10:29 comma-cb80a5fa systemd-timesyncd[3060]: Contacted time server 91.189.91.157:123 (ntp.ubuntu.com).
Sep 18 21:10:29 comma-cb80a5fa systemd-timesyncd[3060]: Initial clock synchronization to Wed 2024-09-18 21:10:29.738480 UTC.
$ sudo mount -o rw,remount /
$ 
```

### DIFF
--- a/Dockerfile.agnos
+++ b/Dockerfile.agnos
@@ -250,8 +250,12 @@ COPY ./userspace/files/apt.conf /etc/apt/apt.conf
 # copy version file
 COPY VERSION /VERSION
 
+# The modification time ("mtime") of this file is used for advancing the system clock in case /var/lib/systemd/timesync/clock does not exist yet
+# Sets initial boot time (as build time) and fixes fstab warning #374 (system time > fstab mtime)
+RUN touch /usr/lib/clock-epoch
+
 # ################## #
-# #### Cleaunup #### #
+# #### Cleanup #### #
 # ################## #
 
 RUN rm -rf /usr/share/icons/* && \
--- a/userspace/usr/comma/fs_setup.sh
+++ b/userspace/usr/comma/fs_setup.sh
@@ -9,6 +9,12 @@ touch /data/etc/timezone
 touch /data/etc/localtime
 mkdir -p /data/etc/NetworkManager/system-connections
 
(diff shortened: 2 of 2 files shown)


adeebshihadeh:
This breaks [this](https://github.com/commaai/openpilot/blob/41a7fa06235cc4ad58921b18d40f199cdf9205bb/common/time.py#L8). 

Aside from that, I don't think we want the time to pick up where it left off. If the time is incorrect, I think I'd prefer it starting from the same point unless you have a good reason for this?

andiradulescu:
@adeebshihadeh was actually asking myself on how openpilot knows that the time is synced or not - now I know

the only reason I thought it's good to pick up from where it left off is that routes/logs are arranged in chronological order, and other events, like sentry, etc. any reasons why starting from the same point is better?

what changes do you see in this PR to be mergeable?

PR #402, opened 2024-11-10
Fix uv installer version to 0.4.30

robin-reckmann:
uv 0.5.0 introduced breaking changes to the installation process. So it's a good idea to fix the uv version in general to keep the agnos build from failing. uv can be updated to 0.5.x in a next step.


adeebshihadeh:
Let's just fix it for 0.5

robin-reckmann:
Closing in favor of #403

PR #406, opened 2024-11-17
Python 3.13

adeebshihadeh:


adeebshihadeh:
pycapnp is the blocker here. probably time to fork pycapnp

https://github.com/capnproto/pycapnp/issues/372

PR #411, opened 2024-11-26
Enable ADB service on startup

amoghmunikote:
PR [#34092](https://github.com/commaai/openpilot/pull/34092)

For bounty [#34041](https://github.com/commaai/openpilot/issues/34041)

Will implement ADB over USB next.


adeebshihadeh:
Done in https://github.com/commaai/agnos-builder/pull/417

Issue #405, opened 2024-11-17
Show dongle ID and other info in tmux bar


adeebshihadeh:
Maybe also temperature? 

`/sys/devices/virtual/thermal/thermal_zone11` is a good one to use for temperature (it's the GPU).

![image](https://github.com/user-attachments/assets/c7c9151c-8961-4eb5-83d7-5d7fa4b6e96c)

Kaweees:
I'm interested in tackling this issue, if it's already not taken

Kaweees:
I'm interested in tackling this issue, if it's already not taken

martinl:
if there is no PR then it is not taken, push one and make it draft to get started

adeebshihadeh:
https://github.com/commaai/agnos-builder/commit/2c089d3665a520a743d9658c4a94ea26bff6e9ed

PR #412, opened 2024-12-04
add GPU temp to tmux info bar

BBBmau:
image

dividing by 1000 is necessary since in unix systems the temperature is given to us in millidegrees (Celsius)

https://developer.toradex.com/software/linux-resources/linux-features/temperature-sensor-linux/?t#nxpfreescale-imx-6-based-modules

seemed helpful to include since it was mentioned in this issue https://github.com/commaai/agnos-builder/issues/405

### DIFF
--- a/userspace/home/.tmux.conf
+++ b/userspace/home/.tmux.conf
@@ -9,7 +9,7 @@ set -g status-left ''
 
 set -g status-left-length 20
 set -g history-limit 7200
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
It's missing a space on the right. Can you also fix up the colors so they're distinct? Feel free to change the dongle color and background too.

image

PR #417, opened 2024-12-07
add adb device agent

BBBmau:
This PR will allow machine to identify the comma device when running `adb devices` on development machine.

Using the following for reference:

- https://docs.thundercomm.com/turbox_doc_demo/documents/common/software/software-release-notes
- https://www.kernel.org/doc/Documentation/usb/gadget_configfs.txt


adeebshihadeh:
Does this work?

BBBmau:
> Does this work?

nope, still dealing with issues where the comma device doesn't appear when plugged in. Only way we can see it listed is if it's set to fastboot mode which shouldn't be necessary for adb.

BBBmau:
the `setup_gadget_mode` script was super helpful. I'm getting the device to appear on USB connection now!

image

adeebshihadeh:
Nice, looks pretty close!

BBBmau:
got it discoverable by adb on host!

It was super minor, once I realized where to set the `99-android.rules` file it was finally able to be discovered.

image

BBBmau:
Moving to ready for review after confirming that `adb shell` works

image

PR #419, opened 2024-12-16
Fix NetworkManager crashes on 24.04

adeebshihadeh:


BBBmau:
Guessing this came from changes that came from the kernel bump

PR #424, opened 2025-01-14
set tty1 in text mode before starting weston

maxime-desroches:


userspace/files/weston.service
```diff
@@ -15,6 +15,7 @@ TTYReset=yes
 TTYVHangup=yes
 TTYVTDisallocate=yes
 
+ExecStartPre=/bin/bash -c "/usr/comma/set_tty_mode.py"
```
adeebshihadeh, line 18: can't we just merge the actual upstream patch into our weston?

PR #433, opened 2025-01-23
fix noisy `push-and-comment` workflow

BBBmau:
noticed the constant emails that I would receive when pushing to my PRs. Did some investigating and found that it was due to `push-and-comment` attempting to download artifacts from `build.yaml`. However we only expect an artifact when the commit message contains `[upload]`

This added check will prevent the job from running if it does not contain `[upload]`

### DIFF
--- a/.github/workflows/push-and-comment.yaml
+++ b/.github/workflows/push-and-comment.yaml
@@ -11,7 +11,8 @@ jobs:
     runs-on: ubuntu-latest
     if: >
       github.event.workflow_run.event == 'pull_request' &&
-      github.event.workflow_run.conclusion == 'success'
+      github.event.workflow_run.conclusion == 'success' &&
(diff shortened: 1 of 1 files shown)


andiradulescu:
This PR makes all `push-and-comment` workflows fail: https://github.com/commaai/agnos-builder/actions/workflows/push-and-comment.yaml
```
[Invalid workflow file: .github/workflows/push-and-comment.yaml#L12](https://github.com/commaai/agnos-builder/actions/runs/12970636514/workflow)
The workflow is not valid. .github/workflows/push-and-comment.yaml (Line: 12, Col: 9): Unrecognized named-value: 'env'. Located at position 116 within expression: github.event.workflow_run.event == 'pull_request' && github.event.workflow_run.conclusion == 'success' && contains(env.LAST_COMMIT_MESSAGE, '[upload]')
```
Let's merge https://github.com/commaai/agnos-builder/pull/437 and see if it works.

PR #438, opened 2025-01-24
Switch to squashfs for tiny rootfs!

adeebshihadeh:
Likely the best solution for https://github.com/commaai/agnos-builder/issues/225

- [x] make RW work (pip/apt/etc). overlay?
- [ ] make sure it doesn't regress boot time
- [ ] find a good balance on compression for build time, download/flash time, and decompression/boot time

---

```
batman@workstation-adeeb:~/agnos-builder$ du -hs output/system.*
3.5G    output/system.img
742M    output/system.squashfs
```

Flashes in 18s over USB 3, no more streaming decompressor in the AGNOS updater or flash.comma.ai, and we can shrink the system partitions 10x for more `userdata` space.

---

* unfortunately, zstd support wasn't introduced until kernel 5.6
* openpilot's updater assumes the image is compressed, so we have to double compress the system image for now
  * @incognitojam after this is merged, you probably want to add an "alt" image that's the direct squashfs image for the flasher. should speed it up a ton


Doncuppjr:
There are at least a few ways to use squashfs. It's a wonderful FS. I wonder where your going with this though. Certainly using squashfs with overlay is going to cut down on NAND wear and tear, but there is the boot time penalty. If you are already compressing the disk image for inet transfer, how does this help?

adeebshihadeh:
I listed the reasons above. The boot time hit isn't great in this PR, but before shipping this, we'll figure out the right tradeoff on size and boot time.

Doncuppjr:
And your kernel has squashfs support built-in, or are you using a cpio wrapper?

adeebshihadeh:
It's built in the kernel. More of the discussion is over at https://discord.comma.ai in the #dev-agnos channel if you wanna join there.

andiradulescu:
@Doncuppjr - reasons using SquashFS for the rootfs of AGNOS:
- reduces image size close to 1GB - as needed here https://github.com/commaai/agnos-builder/issues/225
- reduces the download size without needing compression and (streaming) decompression - both in `openpilot` (when updating) or when flashing with https://flash.comma.ai - thus flashing faster

The kernel has built-in [support for SquashFS](https://github.com/commaai/agnos-kernel-sdm845/commit/6deaba47daa91ec56ed11e6ad252ef46e163dc74).

Of course this comes with drawbacks:
- boot time penalty
- building openpilot freezes the device (running out of memory)

Would be nice if you can join on Discord: [invite](https://discord.comma.ai/) then [#dev-agnos](https://discord.com/channels/469524606043160576/1262118077017882715)

PR #443, opened 2025-01-25
manifest with all partitions

maxime-desroches:
Tested with on device update:
- agnos10 <--> agnos11
- agnos11.x <--> agnos11.x+1

manifest notes:
- `ondevice_hash` is the hash of what is actually written on device during QDL flashing. For sparse image, it's the `FILL` (excluding `FILL` 0) and `RAW` chunks data (no headers). For all other images, it is `hash(image + b'\x00' * ((SECTOR_SIZE - (image_size % SECTOR_SIZE)) % SECTOR_SIZE))`
- `size` for sparse image is the size of the data for the `FILL` and `RAW` chunks only (no headers)
- `hash_raw` for sparse image is the same as `ondevice_hash`

other:
- `agnos.py` currently does not handle `DONT_CARE` sparse chunks, so we assert that none are present for OTA partitions.


scripts/package_ota.py
```diff
@@ -59,19 +97,36 @@ def process_file(fn, name, full_check=True, has_ab=True):
...
+      "num_sectors": num_sectors,
+    }
+
   return ret
 
 
 if __name__ == "__main__":
+  parser = argparse.ArgumentParser(description='Generate a partitions manifest')
+  parser.add_argument('--all-partitions', action='store_true', help='include all partitions, not only the OTA ones')
+  args = parser.parse_args()
```
adeebshihadeh, line 113: this should always generate both manifests together so we always have a matching set

adeebshihadeh:
I propose we also add the userdata caches here:
* flash.comma.ai can also use it
* the full state of what we flash is contained in this manifest

scripts/package_ota.py
```diff
@@ -16,6 +17,48 @@
...
+  GPT(1, 'gpt_main_1', FIRMWARE_DIR / 'gpt_main_1.bin', 0, 6, 'FE8EF7653DB588D7420A625920CA06927DFCB0ED8AFF3E3A1C74A52A24398BA6', False, False, True),
+  GPT(2, 'gpt_main_2', FIRMWARE_DIR / 'gpt_main_2.bin', 0, 6, '5CCFC7240C8CBFA2F1A018A2E376CF274A6BAF858C9BFE71951D8E28CAB53C21', False, False, True),
+  GPT(3, 'gpt_main_3', FIRMWARE_DIR / 'gpt_main_3.bin', 0, 6, 'C707979FA21E89519328F4F30C2B21C9C453401CA8303F914C1873D410A95159', False, False, True),
+  GPT(4, 'gpt_main_4', FIRMWARE_DIR / 'gpt_main_4.bin', 0, 6, 'E9405DCD785DBE79412184E1894A9C51AB7DEB33BB612166C4C42A3D2BF42A0E', False, False, True),
+  GPT(5, 'gpt_main_5', FIRMWARE_DIR / 'gpt_main_5.bin', 0, 6, '21AE965F05B2FA8D02E04F1EB74718F9779864F6EACDEB859757D6435E8CCCE3', False, False, True),
+]
+
+Partition = namedtuple('Partition', ['name', 'path', 'ondevice_hash', 'has_ab', 'ota', 'full_check'])
+PARTITIONS = [
+  Partition('persist', FIRMWARE_DIR / 'persist.bin', '9814B07851292F510F3794B767489F38AB379A99F0EA75DC620AD2D3A496D54D', False, False, True),
```
adeebshihadeh, line 32: why can't we calculate these hashes on the fly?

maxime-desroches: We could, be we would have to duplicate the logic to handle sparse image + the logic for when the image file is not a multiple of 4096 (since those hashes represent what we actually write on the UFS)

scripts/build_dashcam_cache.sh
adeebshihadeh, line 1: probably time to rename this file - build_software_cache?

build_system.sh
```diff
@@ -136,7 +136,7 @@ exec_as_root bash -c "set -e; export ROOTFS_DIR=$ROOTFS_DIR GIT_HASH=$GIT_HASH;
 echo "Unmount filesystem"
 exec_as_root umount -l $ROOTFS_DIR
 
-# Copy system image to output
-cp $ROOTFS_IMAGE $OUT_IMAGE
+# Sparsify system image
+img2simg $ROOTFS_IMAGE $OUT_IMAGE
```
adeebshihadeh, line 140: need both for now since we can't flash this with `edl`. just output a system.img and system.simg

maxime-desroches: `edl` supports sparse images

PR #457, opened 2025-02-12
build `userdata` in CI

maxime-desroches:
Also reduce memory required from 90GB -> 1.8GB

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -60,6 +60,9 @@ jobs:
         GIT_HASH: ${{ github.event.pull_request.head.sha }}
       run: ./build_system.sh
 
+    - name: Build userdata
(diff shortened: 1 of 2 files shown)


andiradulescu:
do we need to build userdata in GitHub CI? although it doesn't sound quite right, we could skip the userdata images in `scripts/package_ota.py`, if they don't exist?

maxime-desroches:
Running locally should always be the same as running CI. The only difference that is ok is the infrastructure that actually runs the test/build.

andiradulescu:
Totally understand. Just that producing artifacts of 5.28 GB is huge. Plus they get uploaded to GitLab LFS when using [upload].

andiradulescu:
can you merge this? I will try to come up with something after.

PR #463, opened 2025-03-11
compile-qtwayland5.sh without checkinstall

andiradulescu:
This fixes the main ui crashing immediately on startup.
```
#0  0x0000000000000000 in ?? ()
#1  0x0000007fb1179500 in ?? () from /usr/lib/aarch64-linux-gnu/qt5/plugins/platforms/libqwayland-egl.so
```

The reason is `checkinstall` doesn't package all the files in `compile-qtwayland5.sh` corectly because of [qmake](https://askubuntu.com/a/1017594/562170), so I replaced it with `dpkg-deb` (which is also used in `build-qt-deb.sh` and `build-libwayland-deb.sh`).

I have no idea how it worked before. Since the `qtwayland5` repo/branch we are using didn't change and neither the `checkinstall` version. Checking `libqwayland-egl.so` on AGNOS v11.10 the file was provided by `qtwayland5` (correctly), so something changed recently.

The `checkinstall` errors were `silent` so this is why the build didn't fail:
```
Error copying ../../../../plugins/platforms/libqwayland-egl.so to /usr/lib/aarch64-linux-gnu/qt5/plugins/platforms/libqwayland-egl.so: Cannot create /usr/lib/aarch64-linux-gnu/qt5/plugins/platforms/libqwayland-egl.so for output
make[4]: [Makefile:611: install_target] Error 3 (ignored)
```

As for `checkinstall` in general, it doesn't have proper maintainers [anymore](https://bugs.launchpad.net/ubuntu/+source/checkinstall) and we should migrate from it. The [proper replacement](https://www.debian.org/doc/manuals/maint-guide/build.en.html) is what is used in https://github.com/commaai/agnos-builder/pull/408 (but not so readable for our usecase). A simpler solution would be to use `dpkg-deb` and store the `DEBIAN/control` files separately (or use another tool to generate them - e.g. `dh_make`).

### DIFF
--- a/userspace/compile-qtwayland5.sh
+++ b/userspace/compile-qtwayland5.sh
@@ -40,13 +40,24 @@ git apply /tmp/agnos/patch-qtwayland-v5.12
 # https://stackoverflow.com/a/75855054/639708
 ln -s libdl.so.2 /usr/lib/aarch64-linux-gnu/libdl.so
 
-mkdir /tmp/build && cd /tmp/build
-qmake /tmp/qtwayland
-
+qmake
 export MAKEFLAGS="-j$(nproc)"
 make
 
-# remove "--fstrans=no" when checkinstall is fixed (still not fixed in 24.04)
-# # https://bugs.launchpad.net/ubuntu/+source/checkinstall/+bug/78455
-checkinstall -yD --install=no --fstrans=no --pkgversion="${VERSION}" --pkgname=qtwayland5 --pkgarch=arm64 --replaces=qtwayland5,libqt5waylandclient5,libqt5waylandcompositor5
-mv qtwayland5*.deb /tmp/qtwayland5.deb
+OUTPUT_DIR=/tmp/qtwayland5
+mkdir $OUTPUT_DIR
+make install INSTALL_ROOT=$OUTPUT_DIR
+
+mkdir $OUTPUT_DIR/DEBIAN
+cat << EOF > $OUTPUT_DIR/DEBIAN/control
+Package: qtwayland5
+Version: ${VERSION}-1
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Thanks!

andiradulescu:
Just checked the differences of `libqwayland-egl.so` between AGNOS with `checkinstall` and created with `dpkg-deb` and noticed that:

checkinstall is stripped / no debug info:
```
file /usr/lib/aarch64-linux-gnu/qt5/plugins/platforms/libqwayland-egl.so
/usr/lib/aarch64-linux-gnu/qt5/plugins/platforms/libqwayland-egl.so: ELF 64-bit LSB shared object, ARM aarch64, version 1 (SYSV), dynamically linked, BuildID[sha1]=d31cfd58953f1a368f2f0002395ab691a632ba91, stripped
```

while the manually built one is not stripped / with debug_info:
```
file /usr/lib/aarch64-linux-gnu/qt5/plugins/platforms/libqwayland-egl.so
/usr/lib/aarch64-linux-gnu/qt5/plugins/platforms/libqwayland-egl.so: ELF 64-bit LSB shared object, ARM aarch64, version 1 (SYSV), dynamically linked, BuildID[sha1]=7afed9e3c98f7b444f0bbd628c15e176d4f3939d, with debug_info, not stripped
```

Also the size is much bigger: Installed-Size: 96.8 MB vs Installed-Size: 8,770 kB.

I will see what's the [easiest way](https://github.com/ruxkor/checkinstall/blob/e48e26acca25db3831c90d4832554eca1f7a172d/checkinstall#L1806) to do this and submit another PR.

PR #466, opened 2025-03-30
ADB: Modified adb.service to auto-configure adb shell properly

sammohana0:
Aim to solve https://github.com/commaai/openpilot/issues/34958
It needs to be tested on comma3x as I don't have one.

### DIFF
--- a/userspace/files/adbd.service
+++ b/userspace/files/adbd.service
@@ -3,4 +3,4 @@ Description=adbd Service
 
 [Service]
 User=root
-ExecStart=/usr/comma/adbd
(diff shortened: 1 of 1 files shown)


userspace/files/adbd.service
```diff
@@ -3,4 +3,4 @@ Description=adbd Service
 
 [Service]
 User=root
-ExecStart=/usr/comma/adbd
+ExecStart=/bin/sh -c 'export TERM=xterm-256color; su - comma -c "cd /data/openpilot && exec /usr/comma/adbd"'
```
maxime-desroches, line 6: This does not work. Also, it should not assume the existence of `/data/openpilot` (for example of first boot)

PR #469, opened 2025-04-16
auto-configure adb shell

rekudyu:
This should auto-configure adb shell as per https://github.com/commaai/openpilot/issues/34958.


rekudyu:
@maxime-desroches can you test this ? thanks to @BBBmau mentioning the stack overflow question this should be working.

adeebshihadeh:
Doesn't work - doesn't appear to do anything

greatgitsby:
there's no `/system/etc/` on agnos

rekudyu:
> there's no `/system/etc/` on agnos

You are right since agnos is based on ubuntu the default shell should is bash. Using bashrc instead of mkshrc should work or else give me output of `echo $SHELL` right after using `adb shell` command

greatgitsby:
adeeb addressed the primary issue: https://github.com/commaai/openpilot/pull/35135

probably can close this PR, i took a stab at it myself and adb shell is almost impossible to configure

PR #476, opened 2025-06-24
Avahi ssh publish

spektor56:
Add avahi to publish the SSH service on the network.

discovery is in the format:
Name: "openpilot SSH - $MODEL - [$HOSTNAME]"
type: _ssh._tcp
port: 22
txt-record: vendor=comma
txt-record: device=openpilot
txt-record: model=$MODEL


userspace/files/avahi-override.conf
```diff
@@ -0,0 +1,3 @@
+[Unit]
+After=network.target
+Wants=network.target
```
adeebshihadeh, line 3: why do you need the override?

userspace/files/avahi-ssh-publish.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Publish commaai SSH service via Avahi
```
adeebshihadeh, line 2:
```suggestion
Description=Publish comma SSH service via Avahi
```

userspace/usr/comma/avahi-publish-service.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+# Read the raw device model
+RAW_MODEL=$(tr -d '\0' < /sys/firmware/devicetree/base/model)
+
+# Extract the device model (e.g., tici, tizi, mici)
+MODEL=$(echo "$RAW_MODEL" | sed 's/comma //g' | awk '{print $1}')
+
+[ -z "$MODEL" ] && MODEL="comma-unknown"
```
adeebshihadeh, line 9:
```suggestion
[ -z "$MODEL" ] && MODEL="unknown"
```

userspace/usr/comma/avahi-publish-service.sh
```diff
@@ -0,0 +1,15 @@
...
+
+HOSTNAME=$(hostname)
+SERVICE_NAME="openpilot SSH - $MODEL - [$HOSTNAME]"
+
+# Publish avahi service
+exec avahi-publish -s "$SERVICE_NAME" _ssh._tcp 22 "vendor=comma" "device=openpilot" "model=$MODEL"
```
adeebshihadeh, line 15: just use "comma". no "openpilot"

adeebshihadeh (changes requested):

userspace/files/avahi-override.conf
spektor56: override is needed because "set-hostname.sh" is called by the serial-hostname.service.  That service has the requirement "Before=network.target".  I could put "After=serial-hostname.service" but i didn't really want to create a dependency on that service existing.  Without this code, avahi starts when the hostname is still "comma" and not "comma-serial".  Let me know what you want to do here, if you want me to leave a comment in the code or make a change.

spektor56:
Thinking of adding the device alias from connect as one records, assuming that's stored in params...

spektor56:
Added device alias from connect to a txt-record, ready for review

PR #477, opened 2025-07-15
Remove deprecated pkla files

sshane:
Partially closes https://github.com/commaai/openpilot/issues/35707, there is another leak that occurs even with no rules active.

polkitd has a compat rule file for legacy pkla files which we use, however there appears to be a leak of a few KB when [spawning a new process](https://www.freedesktop.org/software/polkit/docs/0.106/polkit.8.html), which is what the compat rule does.

I ran `HARDWARE.get_modem_temperatures()` at ~10-20 Hz (as fast as possible), normally `hardwared` runs this every 10 seconds. Memory usage for `/usr/lib/polkit-1/polkitd --no-debug`:
- master AGNOS: 32,444 KB (21:29 min)
- this PR: 54,176 KB (20:00 min) ???

```js
comma@comma-50194f77:/usr/share/polkit-1/rules.d$ cat 49-polkit-pkla-compat.rules 
polkit.addAdminRule(function(action, subject) {
        //polkit.log('Starting pkla-admin-identities\n');
        // Let exception, if any, propagate to the JS authority
        var res = polkit.spawn(['/usr/sbin/pkla-admin-identities']);
        //polkit.log('Got "' + res.replace(/\n/g, '\\n') + '"\n');
        if (res == '')
                return null;
        var identities = res.split('\n');
        //polkit.log('Identities: ' + identities.join(',') + '\n');
        if (identities[identities.length - 1] == '')
                identities.pop()
        //polkit.log('Returning: ' + identities.join(',') + '\n');
        return identities;
});

polkit.addRule(function(action, subject) {
        var params = ['/usr/sbin/pkla-check-authorization',
                      subject.user, subject.local ? 'true' : 'false',
                      subject.active ? 'true' : 'false', action.id];
        //polkit.log('Starting ' + params.join(' ') + '\n');
        var res = polkit.spawn(params);                               // <--- LEAK IS HERE
        //polkit.log('Got "' + res.replace(/\n/g, '\\n') + '"\n');
        if (res == '')
                return null;
        return res.replace(/\n$/, '');
});
```

---

Converted from old pklocalauthority files to new rules file.

pkla documentation: https://manpages.ubuntu.com/manpages/focal/man8/pklocalauthority.8.html

polkit rule documentation: https://www.freedesktop.org/software/polkit/docs/latest/polkit.8.html

### DIFF
--- a/Dockerfile.agnos
+++ b/Dockerfile.agnos
@@ -176,8 +176,9 @@ COPY ./userspace/firmware/* /lib/firmware/
 # populate /etc
 COPY ./userspace/files/fstab /etc
 COPY ./userspace/files/profile /etc/profile
-COPY ./userspace/files/allow-network-control.pkla /etc/polkit-1/localauthority/50-local.d/allow-network-control.pkla
-COPY ./userspace/files/allow-modem-control.pkla /etc/polkit-1/localauthority/50-local.d/allow-modem-control.pkla
+#COPY ./userspace/files/allow-network-control.pkla /etc/polkit-1/localauthority/50-local.d/allow-network-control.pkla
+#COPY ./userspace/files/allow-modem-control.pkla /etc/polkit-1/localauthority/50-local.d/allow-modem-control.pkla
+COPY ./userspace/files/comma-polkit.rules /etc/polkit-1/rules.d/
 COPY ./userspace/files/*.rules /etc/udev/rules.d/
 COPY ./userspace/files/ssh*_config /etc/ssh/
 COPY ./userspace/files/logrotate.conf /etc/
--- a/Dockerfile.agnos
+++ b/Dockerfile.agnos
@@ -127,7 +127,6 @@ RUN cd /tmp && \
     libpolkit-gobject-1-0 \
     libsystemd0 \
     polkitd \
-    polkitd-pkla \
     mobile-broadband-provider-info && \
     apt-get -o Dpkg::Options::="--force-overwrite" install -yq ./libqmi.deb && \
     apt-get -o Dpkg::Options::="--force-overwrite" install -yq ./modemmanager.deb && \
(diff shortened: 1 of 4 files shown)


Dockerfile.agnos
```diff
@@ -176,8 +175,7 @@ COPY ./userspace/firmware/* /lib/firmware/
 # populate /etc
 COPY ./userspace/files/fstab /etc
 COPY ./userspace/files/profile /etc/profile
-COPY ./userspace/files/allow-network-control.pkla /etc/polkit-1/localauthority/50-local.d/allow-network-control.pkla
-COPY ./userspace/files/allow-modem-control.pkla /etc/polkit-1/localauthority/50-local.d/allow-modem-control.pkla
+COPY ./userspace/files/comma-polkit.rules /etc/polkit-1/rules.d/
```
sshane, line 178: Nothing is in `/etc/polkit-1/rules.d/`, all existing rules are in `/usr/share/polkit-1/rules.d`, so using that instead

sshane:
So this PR stops the memory from growing by single digit KBsevery single time you run that modem function, however it now grows by hundreds of KB every 10 seconds or so (and this is running the function as fast as possible!). Not sure why this is the result. polkit 126 fixes all memory leaks, so we'll need to update to that instead of this (or as well).

sshane:
Since this removes one memory leak, but introduces another (that is actually worse), going to close. For future reference, this is how to build 126 (takes 30s): https://www.linuxfromscratch.org/blfs/view/systemd/postlfs/polkit.html

sshane:
Needed for https://github.com/commaai/agnos-builder/pull/478

sshane:
Update: this DOES actually solve the memory leak, when running the modem temp function at 0.1Hz as stock. Only if you run it faster than that do you get the higher memory leak result in the PR body!

PR #478, opened 2025-07-15
Limit polkitd memory usage

sshane:
Instead of https://github.com/commaai/agnos-builder/pull/477. Closes https://github.com/commaai/openpilot/issues/35707.

Service will be killed and restarted once it reaches this threshold, which blocks protected (all?) DBus access for only a few seconds every ~[day of uptime](https://github.com/commaai/openpilot/issues/35707#issuecomment-3076052877).


sshane:
This actually won't work reliably on its own. This is because the pkla compat rules we have for network and modem control launch new processes each time: `pkla-check-auth` and `/usr/sbin/pkla-check-authorization comma false false org.freedesktop.ModemManager1.Device.Control`.

When you hit the max memory, systemd doesn't kill the entire unit; rather it kills the newly spawned processes instead. It does *eventually* kill the service and restart it as it should, but I worry that it won't be reliable with a 10 second DBus call rate. At 20Hz it already takes upwards of 10s or more:

```
● polkit.service - Authorization Manager
     Loaded: loaded (/usr/lib/systemd/system/polkit.service; static)
    Drop-In: /etc/systemd/system/polkit.service.d
             └─override.conf
     Active: active (running) since Wed 2025-07-16 01:08:20 UTC; 48s ago    
       Docs: man:polkit(8)
   Main PID: 158082 (polkitd)  
     Memory: 5.5M (max: 5.5M available: 0B)                                                 
     CGroup: /system.slice/polkit.service                                                   
             ├─158082 /usr/lib/polkit-1/polkitd --no-debug
             └─160868 /usr/sbin/pkla-check-authorization comma false false org.freedesktop.ModemManager1.Device.Control

Jul 16 01:09:07 comma-50194f77 polkitd[158082]: Error evaluating admin rules: Error: Helper was signaled with signal SIGKILL (9), stdout=`', stderr=`'
Jul 16 01:09:07 comma-50194f77 polkitd[158082]: Error evaluating admin rules: Error: Helper was signaled with signal SIGKILL (9), stdout=`', stderr=`'
Jul 16 01:09:07 comma-50194f77 polkitd[158082]: Error evaluating admin rules: Error: Helper was signaled with signal SIGKILL (9), stdout=`', stderr=`'
Jul 16 01:09:07 comma-50194f77 polkitd[158082]: Error evaluating admin rules: Error: Helper was signaled with signal SIGKILL (9), stdout=`', stderr=`'
Jul 16 01:09:08 comma-50194f77 polkitd[158082]: Error evaluating admin rules: Error: Helper was signaled with signal SIGKILL (9), stdout=`', stderr=`'
Jul 16 01:09:08 comma-50194f77 polkitd[158082]: Error evaluating admin rules: Error: Helper was signaled with signal SIGKILL (9), stdout=`', stderr=`'
Jul 16 01:09:08 comma-50194f77 polkitd[158082]: Error evaluating admin rules: Error: Helper was signaled with signal SIGKILL (9), stdout=`', stderr=`'
Jul 16 01:09:08 comma-50194f77 polkitd[158082]: Error evaluating admin rules: Error: Helper was signaled with signal SIGKILL (9), stdout=`', stderr=`'
Jul 16 01:09:08 comma-50194f77 polkitd[158082]: Error evaluating admin rules: Error: Helper was signaled with signal SIGKILL (9), stdout=`', stderr=`'
Jul 16 01:09:08 comma-50194f77 polkitd[158082]: Error evaluating admin rules: Error: Helper was signaled with signal SIGKILL (9), stdout=`', stderr=`'
```

sshane:
@adeebshihadeh we can also set a max uptime for the service to restart it, that would be a one line change. I'm a little torn, but I think we do want memory and CPU limiting for other Ubuntu services as you mentioned, so going with this to start with *something* in that vein.

adeebshihadeh:
Agreed

sshane:
This PR is not actually needed, but still good to have! https://github.com/commaai/agnos-builder/pull/477#issuecomment-3076849050

PR #479, opened 2025-07-30
remove empty `persist` image

maxime-desroches:


userspace/files/fstab
```diff
@@ -1,6 +1,6 @@
 /dev/disk/by-label/dsp /dsp auto ro
 /dev/disk/by-partlabel/modem_a /firmware auto ro
-/dev/disk/by-partlabel/persist /persist auto ro,discard,nosuid,nodev,noexec
+/dev/disk/by-partlabel/persist /persist auto ro,discard,nosuid,nodev,noexec,nofail 0 0
```
adeebshihadeh, line 3: what does the `nofail` addition here get us?

maxime-desroches: `systemd` will enter emergency mode on boot if `/dev/disk/by-partlabel/persist` is not valid (for example on first boot where it has been zero-out)

PR #492, opened 2025-10-02
remove armhf from agnos-base.deb

adeebshihadeh:
rest of armhf crap will be deleted with #489


adeebshihadeh:
hmm, might still need this for `adsprpcd`. will test this later

Issue #259, opened 2024-07-25
Speedup system build time


adeebshihadeh:
Goal is to get a <10m build time without a cache on the CI runner. According to https://github.com/commaai/agnos-builder/pull/251, we're currently at 33m. 

For validation, trigger CI 5 times and post the timings of the "[Build system](https://github.com/commaai/agnos-builder/blob/0480a887226779d3f5f1f8394bfaa1afb78f6be0/.github/workflows/build.yaml#L74)" step. The max shouldn't exceed 10m.

1. remove anything that's unnecessary
2. refactoring and reordering steps
3. parallelize steps

If done carefully, it should be possible to do this without a comma 3/3X and have someone on the Discord test it for you at the end.

Emmanueltech:
Hi, I would like to work on this

andiradulescu:
@Emmanueltech you need to show some initial progress in a PR to lock the bounty - the lock then times out after a week of no progress. More info in [Bounties](https://github.com/commaai/openpilot/blob/master/docs/BOUNTIES.md).

robin-reckmann:
@adeebshihadeh I would propose to store the compiled .deb files for capnp, ffpmeg, modemmanager and qtwayland, as well as the wheel for pyqt5 in the repo. This should bring down the docker image creation time by quite a bit. Or do you prefer to compile them every time from source?

adeebshihadeh:
> @adeebshihadeh I would propose to store the compiled .deb files for capnp, ffpmeg, modemmanager and qtwayland, as well as the wheel for pyqt5 in the repo. This should bring down the docker image creation time by quite a bit. Or do you prefer to compile them every time from source?

I suspect it won't be worth it, but feel free to prove me wrong with a PR. Things I'd be concerned about:
* How do I figure out that I have to rebuild `ffmpeg` manually? 
* There's a way to get the build time down enough without making this tradeoff.
* The build process for these checked-in binaries will break since they'll be run rarely.
* As a maintainer, I now have to rebuild these packages for PRs that want to update them (security risk).

andiradulescu:
If we want custom "debs" built and uploaded separately, we should have a comma [PPA](https://launchpad.net/ubuntu/+ppas).

The install of the comma PPA would look like this:
```
sudo add-apt-repository ppa:comma/comma
sudo apt update
```

Then you can just install the packages with `apt`.

Ubuntu launchpad also has it's own CI, so the packages will be built there.

[...]

PR #494, opened 2025-10-13
parallelize some steps

PMQ9:
Try some stuffs for this bounty:
https://github.com/commaai/agnos-builder/issues/259


adeebshihadeh:
This doesn't actually make anything faster, and it's a massive diff for just parallelizing stuff.

Issue #507, opened 2025-11-12
Get Alpine booting on a comma 3X or four


adeebshihadeh:
This bounty is to get alpine booting on our devices to explore switching AGNOS completely over from Ubuntu.

The dream: no more dealing with systemd, we get https://github.com/commaai/agnos-builder/issues/225 for free, and a generally better development experience. Think of the improvement we got from the [Weston removal](https://blog.comma.ai/0101release/#new-ui).

[Alpine](https://www.alpinelinux.org/) seems like a good choice for this, but I'm open to other options too.

 Must work for bounty:
* `apk` must work
* openpilot should mostly work, including graphics and most of the services
* if I SSH in, it should mostly feel like current AGNOS (vim, tmux, etc )
* systemd-related stuff doesn't have to work (NetworkManager, ModemManager, etc)
* a PR where I can build the image myself

If this works and feels good, we'll setup a bigger bounty for actually getting this shipped!

PR #508, opened 2025-11-13
vibing alpine

adeebshihadeh:
just messing around https://github.com/commaai/agnos-builder/issues/507, this is not a serious attempt


robin-reckmann:
Probably better to base this on a no hf,qt,weston branch

PR #520, opened 2026-02-03
Void

andiradulescu:
https://github.com/commaai/agnos-builder/pull/515


andiradulescu:
`abctl` was not setting the correct success bit thus after a number of restarts was getting unbootable (Failed to boot). Same issue in qdl.js: https://github.com/commaai/qdl.js/pull/120

adeebshihadeh:
Moved to https://github.com/commaai/vamOS!

PR #531, opened 2026-03-03
use new vendored deps from commaai/dependencies

sshane:
split from https://github.com/commaai/agnos-builder/pull/529


adeebshihadeh:
A bunch of these should now be removed from the apt install, otherwise we're installing duplicates now.

sshane:
yeah I'm looking at which we should now, whichever you rm'd in op I was going to here

a test to make sure latest AGNOS runs latest or last release openpilot would be super helpful to prevent this pileup of issues in the future

PR #585, opened 2026-05-04
Update modem fimware

robbederks:
modem fw updated from 00459 -> 00956, potentially fixing use after free bug (commaai/openpilot#35788).

still validating, but looking good so far


robbederks:
This didn't work, had some time where it didn't reproduce, but now it's back at roughly the original rate. See no reason to update.

PR #596, opened 2026-05-06
Speed up "Build system" CI step from ~10min to <5min

### LINKED ISSUE #259: Speedup system build time

Alquh:
Closes #259.

  Targets the "Build system" CI step. No caching introduced - every change
  either removes work or runs existing work in parallel. The total run is
  still validated against a cold runner.

  ## Results

  5 cold runs on `ubuntu-24.04-arm`, "Build system" step times:
  | run | time |
  |---|---|
  | 1 | [4m 29s](https://github.com/Alquh/agnos-builder-build-time/actions/runs/25459962922/job/74699318058) |
  | 2 | [4m 26s](https://github.com/Alquh/agnos-builder-build-time/actions/runs/25460404737/job/74700825370) |
  | 3 | [4m 34s ](https://github.com/Alquh/agnos-builder-build-time/actions/runs/25460890588/job/74702495358) |
  | 4 | [4m 14s](https://github.com/Alquh/agnos-builder-build-time/actions/runs/25461337642/job/74704013174) |
  | 5 | [4m 40s](https://github.com/Alquh/agnos-builder-build-time/actions/runs/25461743801/job/74705372643) |

  Baseline before this PR was ~8m 30s

  ## What changed, by category

  ### 1. Remove unnecessary work
  - Drop `ubuntu-server` from `base_setup.sh`. It pulled in ~100 packages
    (`multipath-tools`, `open-iscsi`, `xfsprogs`, `htop`, `tcpdump`,
    `dnsutils`, `ufw`, `hdparm`, etc.) that AGNOS never uses, and the
    postinst chain dominated build time. The load-bearing pieces it
    provided (`cron`, `netplan.io`, `tmux`, `vim-tiny`) are listed
    explicitly. `services.sh` made tolerant of the resulting missing
    optional units via `disable_q` / `mask_q` helpers.
  - Drop apt packages with no consumer in this repo: `llvm`,
    `libgtk2.0-dev`, `libgdbm-dev`, `libncursesw5-dev`, `apport-retrace`,
    `landscape-common`. `llvm` had been in `base_setup.sh` since the
    initial commit with no reference anywhere; openpilot's `pyproject.toml`
    doesn't pull in `llvmlite`/`numba`.
  - Trim diagnostic-only "nice to haves" from `install_extras.sh`: `btop`,
    `hyperfine`, `ncdu`, `tree`, `wavemon`, `speedtest-cli`, `irqtop`,
    `iperf` (kept `iperf3`). Network/service infra retained.
  - `dpkg` `path-exclude` for `/usr/share/{man,info,doc,groff,lintian,
    linda,locale}` (with `/copyright` and en_US locale preserved). Skips
    the writes for files that ship in nearly every package.
  - Drop `--compile-bytecode` from `uv sync`. .pyc files now compile
    lazily on first import on device. **Tradeoff: slightly slower first
    boot in exchange for image-build time.**

  ### 2. Refactor / reorder
  - Move the `docker buildx build --check` Dockerfile lint passes out of
    `build_system.sh` into a separate `lint-dockerfiles` CI job that runs
    in parallel. Same coverage, no longer on the Build system step's
    critical path.
  - Consolidate the agnos-compiler stage: the libqmi + ModemManager
    compile scripts each ran their own `apt-get update && apt-get install`;
    the union is now installed once in the shared `agnos-compiler` base.
    `meson-install` is also cloned once.
  - Merge apt installs within stages. `base_setup.sh` now sources
    `openpilot_dependencies.sh` and installs both lists in one apt-fast
    invocation. The final stage merges runtime libs + the
    libqmi/modemmanager .debs + extras into one invocation, dropping two
    separate dpkg passes.
  - Replace dead `git.io/vokNn` shortlink for the apt-fast install script
    with the canonical raw GitHub URL (the shortener intermittently
    returns an HTML deprecation page).
  - Enable `dpkg force-unsafe-io` in both compile and base stages — same
    effect as the eatmydata `LD_PRELOAD` trick but using dpkg's own
    supported flag. Skips fsync per package.

 ### 3. Parallelize
  - `agnos-compiler-modemmanager` now `FROM agnos-compiler` instead of
    `FROM agnos-compiler-libqmi`. Buildkit runs both compiles
    concurrently. MM compiles against Ubuntu's `libqmi-glib-dev` (1.34)
    for headers; the runtime image still installs the locally-built 1.36
    `.deb`. libqmi-glib has shipped with `SONAME=5` since libqmi 1.10
    (~2014), so 1.34 → 1.36 is ABI-compatible

  ## Reviewer notes
  - No vendored binaries, no checked-in `.deb`s. Builds remain entirely
    from source on every run.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -11,6 +11,21 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  lint-dockerfiles:
+    name: lint Dockerfiles
+    runs-on: ubuntu-24.04-arm
+    timeout-minutes: 5
+    steps:
+    - uses: actions/checkout@v4
+    - name: Check Dockerfile.agnos
+      run: docker buildx build -f Dockerfile.agnos --check .
+    - name: Check Dockerfile.builder
+      run: |
+        docker buildx build -f Dockerfile.builder --check . \
+          --build-arg UNAME=$(id -nu) \
+          --build-arg UID=$(id -u) \
+          --build-arg GID=$(id -g)
+
   build:
     name: build kernel and system
     runs-on: ubuntu-24.04-arm
--- a/Dockerfile.agnos
+++ b/Dockerfile.agnos
@@ -49,6 +49,7 @@ RUN apt-get update && apt-get install -yq --no-install-recommends \
     libgirepository1.0-dev \
     libglib2.0-dev \
     libgudev-1.0-dev \
+    libmbim-glib-dev \
     libpolkit-gobject-1-dev \
     libqmi-glib-dev \
     libsystemd-dev \
--- a/Dockerfile.agnos
+++ b/Dockerfile.agnos
@@ -27,6 +27,10 @@ EOF
 
 # Common packages + union of libqmi and ModemManager build deps so each
 # compile stage doesn't re-run apt update/install.
+# libqmi-glib-dev (Ubuntu's 1.34) is included so the ModemManager stage
+# can compile in parallel with our libqmi-from-source (1.36) build —
+# libqmi-glib SONAME=5 is ABI-stable across these versions, and the final
+# image installs the locally-built 1.36 .deb for runtime.
 RUN apt-get update && apt-get install -yq --no-install-recommends \
     bash-completion \
     build-essential \
@@ -46,6 +50,7 @@ RUN apt-get update && apt-get install -yq --no-install-recommends \
     libglib2.0-dev \
     libgudev-1.0-dev \
     libpolkit-gobject-1-dev \
+    libqmi-glib-dev \
     libsystemd-dev \
     meson \
     ninja-build \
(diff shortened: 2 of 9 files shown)


Alquh:
@adeebshihadeh would you mind reviewing this when you are able? Thanks!

adeebshihadeh:
This is a huge diff and breaks openpilot (e.g. `apport-retrrace`).

PR #609, opened 2026-07-04
sound: bring up the mici audio card on the mainline kernel

greatgitsby:


userspace/root/usr/comma/comma-init.sh
```diff
@@ -207,7 +207,129 @@ function init_gpio (
...
+
+  local seg jsn
+  ln -sf "$IMG/mba.mbn" "$FARM/qcom/sdm845/mba.mbn"
+  ln -sf "$IMG/modem.mdt" "$FARM/qcom/sdm845/modem_nm.mbn"
+  for seg in "$IMG"/modem.b*; do
+    [[ -e "$seg" ]] || continue
+    ln -sf "$seg" "$FARM/qcom/sdm845/modem_nm.${seg##*/modem.}"
+  done
+
+  ln -sf "$IMG/adsp.mdt" "$FARM/qcom/sdm845/adsp.mbn"
```
adeebshihadeh, line 268: we should actually be able to cutout the ADSP entirely if you want to try that

greatgitsby: i did some cursory research and it seems like an untrod path.. let’s try once mainline is fully working

adeebshihadeh: had a demo going here: https://github.com/commaai/agnos-kernel-sdm845/pull/107

greatgitsby: oh i misunderstood, yea we’re doing something simpler like this on mainline. PR coming up soon

PR #616, opened 2026-08-23
Add read-only USB footage storage

### LINKED ISSUE #38291: Setup mass storage for accessing drives
One more step towards being a good dashcam.

Plug in to your computer, open file explorer, and drag any footage you want off.
* Can we make this work on Windows, macOS, and Linux without needing to install drivers?
* This unfortunately requires 2x cables like flashing, just inserted in the other order.
* Should we expose the whole `userdata` partition or just `realdata`?

There are some natural follow-ups to this, such as making the on-disk format match the good dashcams.

Amacfa:
## Summary

Adds driverless, read-only USB Mass Storage access to completed openpilot footage on comma 3X/four.

Fixes commaai/openpilot#38291 ($300 bounty).

The implementation is ready for review. A comma-assigned storage-only USB identity and Windows validation remain acceptance blockers.

## Approach

The live userdata partition remains mounted as ext4 and is never exposed to the host. A new system service instead:

- waits for a normal power-first USB host connection;
- selects completed, stable `realdata` segments and rejects active or unsafe trees;
- hard-links selected files into a private per-session snapshot on the same filesystem;
- synthesizes a FAT32 disk from that snapshot with Ubuntu Noble's `nbdkit-floppy` and COW filter;
- repairs the small set of nbdkit 1.36 FAT32 metadata fields needed by native host readers; and
- attaches the resulting regular file to a removable, read-only configfs mass-storage LUN.

The hard-link snapshot freezes the exported namespace and keeps an inode readable if loggerd's deleter unlinks its original name. It does not make file contents immutable, so `.lock` exclusion, an age threshold, two matching scans, and a final pre-export layout check enforce the completed-file invariant. A free-space guard 1 GiB / 1 percentage point above loggerd's 5 GiB / 10% floor ejects and removes the snapshot before pinned files can exhaust userdata.

The gadget lifecycle is serialized through a root-owned lock. Teardown is ordered as LUN clear, FUSE unmount, exporter stop, then snapshot cleanup. A post-stop recovery path removes only an exactly identified dead `nbdfuse` mount and managed snapshot after systemd has killed the service cgroup. Host eject, physical detach, forced service stop, stale runtime state, `PREVENT MEDIUM REMOVAL`, and low-space events fail closed.

## USB integration

- Refactors `set_adb.sh` through one serialized configfs helper.
- With ADB enabled, preserves NCM at MI_00/01 and ADB at MI_02, then appends mass storage at MI_03 under the existing identity.
- With ADB disabled, exposes mass storage only. This personality intentionally refuses to bind until comma supplies a separate owner-approved VID/PID, avoiding Windows driver-cache collisions from reusing interface IDs under `04d8:1234`.
- Adds `usb-storage.service`, ordered shutdown handling, required Noble packages, root-owned lock creation, and service enablement.
- Re-enumerates only after validating the exact requested personality, single config, single LUN, backing path, and read-only policy. Unexpected configurations or LUNs are preserved but left unbound.

## Safety and privacy boundaries

- Exports `realdata` only; no params, credentials, openpilot checkout, or raw userdata blocks.
- Host writes are rejected by the kernel LUN (`ro=1`); the backing export is COW-isolated.
- Segments with active `.lock` files, unstable metadata, symlinks, special files, unsafe FAT/Windows names, case/NFC collisions, or FAT32-oversized files are excluded.
- Empty artifacts are omitted because Noble nbdkit 1.36 emits an invalid FAT chain for them; usable sibling footage remains available.
- Historical timestamp, current log-ID, and canonical/copied route-directory forms are supported. Legacy `|` is mapped to openpilot's Windows-safe `_` representation in the snapshot only.
- No live-control, panda safety, vehicle, camera, bootloader, or partition behavior is changed.

## Validation

Software:

- `python3 -m unittest discover -s userspace/tests -p 'test_*.py' -v` — 125/125 passed on the clean PR tree.
- Ruff, `py_compile`, shell syntax, and `git diff --check` — passed.
- The ARM workflow runs the production snapshot/layout/repair code against installed Noble `nbdkit 1.36.3`, then verifies padded and natural-size images with `fsck.fat`, mtools extraction, matching hashes, and a multi-cluster FAT root containing 172 route directories.
- Upstream PR run [`32665270834`](https://github.com/commaai/agnos-builder/actions/runs/32665270834) — passed 125/125 tests, padded/natural/multi-cluster-root installed-Noble FAT integration, kernel build, full system build, and filesystem statistics at exact head `5ca14c2`.

comma four bench validation, without flashing:

- Kept the device on its original successful AGNOS 18.4 slot. The built target root was mounted read-only with `norecovery` and used as a chroot on the common device kernel; no updater, A/B write, camera node, real route, or vehicle was used.
- macOS: 20/20 native enumerate, FAT32 browse, six-file copy/hash, write-denial, eject, and re-enumeration cycles passed. The 19 between-cycle disconnects were deliberate configfs disconnects, not physical cable pulls.
- Ubuntu 24.04 ARM: 1/1 native enumerate, UTF-8 FAT32 browse, six-file copy/hash, write-denial, and eject passed after a real port-1 unplug/replug while port 2 remained powered.
- Final cleanup restored the stock AGNOS 18.4 NCM+ADB gadget. No stale LUN, FUSE mount, or exporter process remained, and no ext4/journal/kernel error was observed.

## Acceptance status / blockers

- [x] comma four hardware path exercised with port 2 powered before port 1 host data.
- [x] macOS uses its inbox storage/FAT driver and copies verified files.
- [x] Linux uses its inbox storage/FAT driver and copies verified files.
- [x] Host write access is denied.
- [x] Repeated macOS sessions and one physical Linux reconnect/eject recover cleanly.
- [ ] Windows native enumeration and File Explorer copy still need to be run.
- [ ] A distinct owner-approved storage-only VID/PID is required before ADB-off mode can bind.
- [ ] The exact target image was not booted; validation used its userspace read-only on the live common kernel, so target systemd boot ordering is not claimed.

## Known limitations

- FAT32 cannot represent an individual file of 4 GiB or larger; a segment containing one is excluded. Normal one-minute route files are far below this limit.
- The volume is a point-in-time read-only view built at host attachment. New segments appear after the next physical reconnect.
- Zero-byte files contain no retrievable payload and are omitted due to the Noble plugin bug described above.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -21,10 +21,20 @@ jobs:
         lfs: true
         submodules: true
 
+    - name: Install USB storage integration dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y --no-install-recommends dosfstools libnbd-bin mtools nbdkit
+
     - name: Test userspace helpers
       env:
         PYTHONDONTWRITEBYTECODE: "1"
       run: python3 -m unittest discover -s userspace/tests -p 'test_*.py' -v
+
+    - name: Test installed nbdkit FAT export
+      env:
+        PYTHONDONTWRITEBYTECODE: "1"
+      run: python3 userspace/tests/integration_usb_storage_nbdkit.py
 
     - name: Get commit message
       run: |
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -21,6 +21,11 @@ jobs:
         lfs: true
         submodules: true
 
+    - name: Test userspace helpers
+      env:
+        PYTHONDONTWRITEBYTECODE: "1"
+      run: python3 -m unittest discover -s userspace/tests -p 'test_*.py' -v
+
     - name: Get commit message
       run: |
         {
--- a/userspace/base_setup.sh
+++ b/userspace/base_setup.sh
@@ -30,6 +30,7 @@ apt-fast install --no-install-recommends -yq \
     cpuset \
     dnsmasq-base \
     evtest \
+    fuse3 \
     git \
     git-core \
     git-lfs \
@@ -55,12 +56,14 @@ apt-fast install --no-install-recommends -yq \
     libgtk2.0-dev \
     libi2c-dev \
     libncursesw5-dev \
+    libnbd-bin \
     libnss-myhostname \
     libqmi-utils \
     libssl-dev \
     locales \
     llvm \
     nano \
+    nbdkit \
     net-tools \
     nload \
     network-manager \
--- a/userspace/root/usr/comma/cleanup_usb_storage.py
+++ b/userspace/root/usr/comma/cleanup_usb_storage.py
@@ -1,0 +1,190 @@
+#!/usr/bin/env python3
+"""Remove only a stopped USB storage service's narrowly owned artifacts."""
+
+from __future__ import annotations
+
+import argparse
+import os
+from pathlib import Path
+import stat
+import subprocess
+import sys
+from collections.abc import Callable, Sequence
+from typing import Any
+
+from usb_storage import MANAGED_MOUNT_DIRECTORY, MANAGED_MOUNT_NAME, SnapshotBuilder, StorageError
+
+
+def _decode_mountinfo_field(value: str) -> str:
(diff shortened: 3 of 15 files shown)


adeebshihadeh:
Looks like LLM slop

